PR #363, opened 2024-06-18
e2e: Github Actions workflow

trzysiek:


trzysiek:
Implemented in a better and full way by @mieciu

PR #386, opened 2024-06-22
Attempt to unify sampler

jakozaur:
I try to refactor and see if we can fold facets into general aggregation. This is not working fully, but feel that we are close.

Missing pieces are:
- [ ] We generated too many SQL and didn't fold them into one.
- [ ] We can't have an easy way to re-use some results (e.g. a group of facet numeric to count min, max, etc. on them)
- [ ] We can't generate one out of bucket aggregation buckets.

Would need to talk to @trzysiek to see if this is solveable.


jakozaur:
This was replaced by pancake SQLs.

PR #507, opened 2024-07-09
Quickfix: `must_not` generated incorrect SQL

trzysiek:
Simple error, we translated `must_not (a && b)` into `NOT a OR NOT b` which of course isn't correct. Changed it to `not (a or b)`.
Before (we should have 0 results as all `@timestamp`s fall into second `range`):
Cursor_and_Console_-_Dev_Tools_-_Elastic
After:
Screenshot 2024-07-10 at 00 36 03

Proof that it's indeed not a AND not b in Elastic: with one condition, we have >0 results, with 2 we have none, so it can't be OR like we used to have.
Screenshot 2024-07-10 at 10 36 15
Screenshot 2024-07-10 at 10 36 23

### DIFF
--- a/quesma/queryparser/query_parser.go
+++ b/quesma/queryparser/query_parser.go
@@ -427,7 +427,10 @@ func (cw *ClickhouseQueryTranslator) iterateListOrDictAndParse(queryMaps interfa
 		return cw.parseQueryMapArray(queryMapsTyped)
 	case QueryMap:
 		simpleQuery := cw.parseQueryMap(queryMapsTyped)
-		return []model.Expr{simpleQuery.WhereClause}, simpleQuery.CanParse
+		if simpleQuery.WhereClause != nil {
+			return []model.Expr{simpleQuery.WhereClause}, simpleQuery.CanParse
+		}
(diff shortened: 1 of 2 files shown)


quesma/queryparser/query_parser.go
```diff
@@ -427,7 +427,10 @@ func (cw *ClickhouseQueryTranslator) iterateListOrDictAndParse(queryMaps interfa
 		return cw.parseQueryMapArray(queryMapsTyped)
 	case QueryMap:
 		simpleQuery := cw.parseQueryMap(queryMapsTyped)
-		return []model.Expr{simpleQuery.WhereClause}, simpleQuery.CanParse
+		if simpleQuery.WhereClause != nil {
+			return []model.Expr{simpleQuery.WhereClause}, simpleQuery.CanParse
+		}
+		return []model.Expr{}, simpleQuery.CanParse
```
trzysiek, line 433: Not sure why but after my fix a few tests started panicking (this `WhereClause` was a `null`, and some highlighter visitor panicked when it saw this `null`). I think this change makes sense

jakozaur (approved):

PR #508, opened 2024-07-10
Add `_bulk` batch size logging

mieciu:
As we've found out, the size of a bulk ingest payload is crucial to the ingest performance.

Hence, we should log this piece of information.

### DIFF
--- a/quesma/quesma/functionality/bulk/bulk.go
+++ b/quesma/quesma/functionality/bulk/bulk.go
@@ -26,7 +26,9 @@ func Write(ctx context.Context, defaultIndex *string, bulk types.NDJSON, lm *cli
 	cfg config.QuesmaConfiguration, phoneHomeAgent telemetry.PhoneHomeAgent) (results []WriteResult) {
 	defer recovery.LogPanic()
 
-	indicesWithDocumentsToInsert := make(map[string][]types.JSON, len(bulk))
+	bulkSize := len(bulk)
+	logger.Info().Msgf("Processing %d documents in _bulk", bulkSize/2)
+	indicesWithDocumentsToInsert := make(map[string][]types.JSON, bulkSize)
 
 	err := bulk.BulkForEach(func(op types.BulkOperation, document types.JSON) {
 


quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -26,7 +26,10 @@ func Write(ctx context.Context, defaultIndex *string, bulk types.NDJSON, lm *cli
 	cfg config.QuesmaConfiguration, phoneHomeAgent telemetry.PhoneHomeAgent) (results []WriteResult) {
 	defer recovery.LogPanic()
 
-	indicesWithDocumentsToInsert := make(map[string][]types.JSON, len(bulk))
+	bulkSize := len(bulk)
+	logger.Info().Msgf("Inserting %d documents using _bulk", bulkSize/2)
```
pivovarit, line 30: Debug?

mieciu: I'd leave it super visible, at least for now.
In case someone hits poor ingest performance, that'd be our top of mind to check for, so I would prefer having this information really apparent to the user and not require re-setting the log level.

Plus, we get to see that in telemetry 😉

jakozaur: This could be very noisy, I would add some throttling... zero logger got infra for that.

jakozaur: https://github.com/rs/zerolog?tab=readme-ov-file#log-sampling

jakozaur (approved):

jakozaur (approved):

quesma/quesma/functionality/bulk/bulk.go
mieciu: For now keeping it simple with https://github.com/QuesmaOrg/quesma/pull/508/commits/579af88a16e70fbdeb2f22c0534fe4e28c906ec4 
Unfortunately zero log's sampler couldn't be attached to our existing logger somehow, ditching all the logging setup we have.

PR #511, opened 2024-07-10
Make SQL queries clickable in Live Tail Drilldown

avelanarius:
After this change each SQL query in the "Translated SQL" view on Live Tail Drilldown page is now clickable and routes to the ClickHouse `/play` webpage with the query pre-filled-in, which makes it very convenient to quickly run and modify it.

This change required changing the way query bodies are stored internally - previously the queries were concatenated to a single large blob, but now each query is stored separately (as a `[]byte`), making it possible to display them separately in the UI.


avelanarius:
A quick video how it looks like:

https://github.com/QuesmaOrg/quesma/assets/10003183/7bd1b771-0819-4ce8-9252-533c15e710d8

pivovarit (approved):
Nice!

PR #518, opened 2024-07-11
Simplify json rendering refactor

jakozaur:
Another stepping stone toward unifying rendering json:
- Make JSON rendering step more consistent, stop at the same level
- the logic is much shorter
- do not pass array with one element

@trzysiek should review


pivovarit (approved):

trzysiek (approved):
very cool, simplifies things

PR #520, opened 2024-07-12
Rewrite _resolve endpoint to use the schema registry

avelanarius:
Before this change, the `_resolve` endpoint had several bugs. One of them, which affected me, was wrong result for an index present in Elastic, but missing from ClickHouse and not enabled in the config. In that case Quesma should reach to Elastic, but previously it just looked at ClickHouse and immediately returned 404.

Since the previous logic was quite convoluted and used older `LogManager` instead of `schema.Registry`, this PR largely rewrites the implementation to be simpler, better documented and to use the `schema.Registry`.

I extensively tested all possible scenarios (see the table in `Test_combineSourcesFromElasticWithRegistry`), both as a unit test and manually in Kibana. 

The resolve code is now also moved to `functionality/resolve` package.

To make it possible to use the `schema.Registry` in the new code, `schema.Schema` had to be extended with new boolean flag `ExistsInDataSource`. Before the addition of this boolean flag, schema registry contained both: schemas without created table in ClickHouse (only specified/enabled in configuration) and schemas with created table in ClickHouse - no easy way to differentiate between them.

### DIFF
--- a/quesma/quesma/router.go
+++ b/quesma/quesma/router.go
@@ -15,6 +15,7 @@ import (
 	"quesma/quesma/functionality/bulk"
 	"quesma/quesma/functionality/doc"
 	"quesma/quesma/functionality/field_capabilities"
+	"quesma/quesma/functionality/resolve"
 	"quesma/quesma/functionality/terms_enum"
 	"quesma/quesma/mux"
 	"quesma/quesma/routes"
@@ -24,7 +25,6 @@ import (
 	"quesma/telemetry"
 	"quesma/tracing"
 	"regexp"
-	"slices"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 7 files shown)


quesma/quesma/router_test.go
```diff
@@ -181,3 +183,97 @@ func Test_matchedAgainstBulkBody(t *testing.T) {
 		})
 	}
 }
+
+func Test_combineSourcesFromElasticWithRegistry(t *testing.T) {
```
avelanarius, line 187: `router_test.go` is probably not the best place for this test, similarly for `combineSourcesFromElasticWithRegistry` in `router.go`.

Should I move it somewhere else? (where?)

quesma/quesma/router.go
```diff
@@ -84,75 +84,47 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
...
+						{
+							Name:           normalizedPattern,
+							BackingIndices: []string{normalizedPattern},
+							TimestampField: `@timestamp`,
+						},
+					},
+				}), nil
 			}
+
+			// ...index not found in schema.Registry (meaning the user did not configure it), so let's fall back to
```
pivovarit, line 114: I think "fall back" term might be slightly misleading - we don't really "fall back" because the first one did not work - we simply router to Elastic because no one explicitly stated that this particular index name should be resolved against Clickhouse

quesma/quesma/router_test.go
pivovarit: I'd create a dedicated file for `_resolve`-related tests

quesma/quesma/router.go
```diff
@@ -84,75 +84,47 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
...
 
-				return resolveIndexResult(sources), nil
+				return resolveIndexResult(elasticsearch.Sources{
+					Indices: []elasticsearch.Index{},
+					Aliases: []elasticsearch.Alias{},
+					DataStreams: []elasticsearch.DataStream{
+						{
+							Name:           normalizedPattern,
+							BackingIndices: []string{normalizedPattern},
+							TimestampField: `@timestamp`,
```
pivovarit, line 108: That's something we'll need to improve in the future - we can't hardcode it. But before we can fix it, we'll need to expand schema to carry information about main ordering property of each index

quesma/quesma/router.go
```diff
@@ -401,7 +373,36 @@ func elasticsearchInsertResult(body string, statusCode int) *mux.Result {
...
+				BackingIndices: []string{indexName},
+				TimestampField: `@timestamp`,
+			})
+		}
+	}
+
+}
+
 func resolveIndexResult(sources elasticsearch.Sources) *mux.Result {
+	if len(sources.Aliases) == 0 && len(sources.DataStreams) == 0 && len(sources.Indices) == 0 {
```
pivovarit, line 402: No need to check `len(sources.Aliases)` - they are just pointers to existing data structures

quesma/quesma/router_test.go
avelanarius: OK, what about the implementation itself (currently in `router.go`)? Should it be in `resolve.go` or maybe somewhere in `schema.go`, `registry.go`?

pivovarit: Ideally, the functionality should be pulled into a dedicated package and live alongside those:
image

The router should merely just use the exported functionality of a new package

avelanarius: Moved the code and the test to `functionality/resolve` package.

quesma/quesma/router.go
avelanarius: Writing it this way makes it more obvious to the reader that the logic is: "empty response results in 404"

quesma/quesma/router.go
avelanarius: Changed the wording to `...index not found in schema.Registry (meaning the user did not configure it), but it could exist in Elastic`

avelanarius:
v2:
- rebased on current `main`, dealt with conflicts
- addressed review comments
- moved the `resolve` code and test code to a separate file/package

jakozaur (approved):
Great work! Still some things might change after we enable new defaults, but we are going in right direction.

PR #521, opened 2024-07-12
Add optimization pipeline

nablaone:
Let's introduce the SQL query optimization pipeline.

Now there are only two optimizations supported: 
- truncate dates to 5 minutes, it  will limit how many distinct queries we perform
- cache aggregation queries ("GROUP BY"), these queries can be slow, and produce a low number of rows, 

Optimizations are disabled for tests. 
Currently,  performed optimizations are not visible in the "Live View" tab. I will add it in the next PR.

### DIFF
--- a/quesma/clickhouse/quesma_communicator.go
+++ b/quesma/clickhouse/quesma_communicator.go
@@ -136,7 +136,6 @@ func executeQuery(ctx context.Context, lm *LogManager, query *model.Query, field
 	settings["allow_ddl"] = "0"
 
 	if query.OptimizeHints != nil {
-		fmt.Println("Applying performance settings", query.OptimizeHints.Settings)
 		for k, v := range query.OptimizeHints.Settings {
 			settings[k] = v
 		}
--- a/quesma/clickhouse/quesma_communicator.go
+++ b/quesma/clickhouse/quesma_communicator.go
@@ -72,7 +72,7 @@ func (lm *LogManager) ProcessQuery(ctx context.Context, table *Table, query *mod
 
 	}
 
-	rows, err := executeQuery(ctx, lm, query.SelectCommand.String(), columns, rowToScan)
+	rows, err := executeQuery(ctx, lm, query, columns, rowToScan)
 
 	if err == nil {
 		for _, row := range rows {
@@ -121,8 +121,10 @@ func (lm *LogManager) explainQuery(ctx context.Context, query string, elapsed ti
 	}
 }
 
-func executeQuery(ctx context.Context, lm *LogManager, queryAsString string, fields []string, rowToScan []interface{}) ([]model.QueryResultRow, error) {
+func executeQuery(ctx context.Context, lm *LogManager, query *model.Query, fields []string, rowToScan []interface{}) ([]model.QueryResultRow, error) {
 	span := lm.phoneHomeAgent.ClickHouseQueryDuration().Begin()
+
+	queryAsString := query.SelectCommand.String()
 
 	// We drop privileges for the query
 	//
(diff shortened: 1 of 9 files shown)


quesma/clickhouse/quesma_communicator.go
```diff
@@ -133,6 +135,15 @@ func executeQuery(ctx context.Context, lm *LogManager, queryAsString string, fie
 	settings["readonly"] = "1"
 	settings["allow_ddl"] = "0"
 
+	if query.OptimizeHints != nil {
+		fmt.Println("Applying performance settings", query.OptimizeHints.Settings)
```
jakozaur, line 139: Change to proper log line?

quesma/optimize/cache_group_by.go
```diff
@@ -0,0 +1,37 @@
...
+//  SYSTEM DROP QUERY CACHE
+//
+
+type cacheGroupByQueries struct {
+}
+
+func (s *cacheGroupByQueries) Transform(queries []*model.Query) ([]*model.Query, error) {
+
+	for _, query := range queries {
+
```
jakozaur, line 28: FIY: It would be nice if customer could write hints like that.

jakozaur (commented):
Great job, though!
1. Let's make this configurable from YAML.
I struggle a little with this concept:
2. We should only truncate when the time range is wide and the query is more than one day. Otherwise, you will deal with a lot of quirks. Also only do that for group by searches.
3. Currently truncationtion can change time range length, we should take into account gt or lt and always select previous tick, next tick.

quesma/clickhouse/quesma_communicator.go
nablaone: It's not necessary here.

quesma/optimize/cache_group_by.go
nablaone: Just like any other transformer.

nablaone:
1. Let's add in the next PR. 

2,3 Solved in the latest commit.

quesma/model/query.go
```diff
@@ -11,10 +11,18 @@ const (
 	noLimit             = 0
 )
 
+// QueryOptimizeHints contains hints for query execution, e.g. peformance settings, temparaty table usage
```
pivovarit, line 14:
```suggestion
// QueryOptimizeHints contains hints for query execution, e.g., performance settings, temporary table usage
```

quesma/quesma/quesma.go
```diff
@@ -126,6 +126,11 @@ func NewHttpProxy(phoneHomeAgent telemetry.PhoneHomeAgent, logManager *clickhous
 
 	queryRunner.DateMathRenderer = queryparser.DateMathExpressionFormatLiteral
 
+	// TODO this should be configurable somehow
```
pivovarit, line 129: Our YAML file ideally

quesma/quesma/quesma.go
```diff
@@ -126,6 +126,11 @@ func NewHttpProxy(phoneHomeAgent telemetry.PhoneHomeAgent, logManager *clickhous
...
+	//
+	// tests should not be run with optimization enabled by default
+	queryRunner.EnableQueryOptimization()
```
jakozaur, line 132:
```suggestion
        // TODO: Enable it in YAML
	// queryRunner.EnableQueryOptimization()
```

jakozaur (changes requested):
I really like this change, but struggle to enable something by default that could break semantic. Let's disable it by default merge and add it in next PR./

quesma/optimize/trunc_date.go
```diff
@@ -0,0 +1,253 @@
...
+package optimize
+
+import (
+	"fmt"
+	"quesma/model"
+	"strings"
+	"time"
+)
+
+type truncateDateVisitor struct {
```
pdelewski, line 12: Please add comment about the way of working

quesma/optimize/trunc_date.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (v *truncateDateVisitor) VisitLiteral(e model.LiteralExpr) interface{} {
+	return e
+}
+
+func (v *truncateDateVisitor) VisitInfix(e model.InfixExpr) interface{} {
+
+	op := strings.ToLower(e.Op)
+	if op == "and" {
```
pdelewski, line 95: I think that body of this if branch could be extracted as separate function and rewritten a bit to reverse conditions,
```
  if op != "and" {
     return
  }
```

nablaone: Yep. I've extracted it to a separate method.

jakozaur (approved):

PR #522, opened 2024-07-12
Json rendering refactor

jakozaur:
Changes to render logic:
1. Unify the whole JSON logic with render inside `TranslateSqlResponseToJson`:
- this allows to generate buckets as well as top-level values
- we always return `model.JsonMap` and no longer assume one element array
2. As a proof of concept, I implement `doc_count_error_upper_bound` for `terms` in new architecture.
3. Query translation logic got much simpler.
4. Need to do a lot of tiny fixes.

### DIFF
--- a/quesma/model/query.go
+++ b/quesma/model/query.go
@@ -32,8 +32,8 @@ type (
 	QueryType interface {
 		// TranslateSqlResponseToJson 'level' - we want to translate [level:] (metrics aggr) or [level-1:] (bucket aggr) columns to JSON
 		// Previous columns are used for bucketing.
-		// For 'bucket' aggregation result is a slice of buckets, for 'metrics' aggregation it's a single bucket (only look at [0])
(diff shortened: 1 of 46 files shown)


quesma/model/query.go
```diff
@@ -32,8 +32,8 @@ type (
 	QueryType interface {
 		// TranslateSqlResponseToJson 'level' - we want to translate [level:] (metrics aggr) or [level-1:] (bucket aggr) columns to JSON
 		// Previous columns are used for bucketing.
-		// For 'bucket' aggregation result is a slice of buckets, for 'metrics' aggregation it's a single bucket (only look at [0])
-		TranslateSqlResponseToJson(rows []QueryResultRow, level int) []JsonMap
+		// For 'bucket' aggregation result is a map wrapperd in 'buckets' key.
```
pivovarit, line 35:
```suggestion
		// For 'bucket' aggregation result is a map wrapped in 'buckets' key.
```

trzysiek (approved):
nice! all looks good

PR #523, opened 2024-07-12
optimizer: merging metrics aggregations

trzysiek:
I'll update the code after Rafał's merge
It basically seems to work in all cases except 2 specific aggregations, which have > 1 column, unlike others - need a simple fix for that


jakozaur:
Closed in favor of pancakes SQL.

PR #530, opened 2024-07-15
Enable/disable optimizer via configuration

nablaone:
We may enable/disable each optimizing transformer on global or index scope.

Here is a configuration snippet that enables/disables existing transformers:
```
# enable on the global scope
optimizers:
  cache_group_by_queries: true
  truncate_date: true

# enable on index scope
indexes:
  some-index:
    enabled: true
    optimizers:
       cache_group_by_queries: true
       truncate_date: true
```


jakozaur (approved):
Ideally would love to specify by how much date should be truncated, but fine to merge as is for now.

PR #531, opened 2024-07-15
Base model.Expr visitor 

nablaone:
We introduce a base visitor that implements the `model.ExprVisitor`.  We usually implement only one or two methods for the visitor. Other methods are the same. 

@pdelewski please take a look at this


pdelewski:
Look great to me :)

pdelewski (approved):

PR #532, opened 2024-07-15
POC N index -> 1 table mapping

pdelewski:
This PR adds new configuration settings to map input/source indexes to internal database table (that's proposal, I'm open to suggestions)
```
indexMappings:
  big_kibana_common_table:
    sourceIndexes: ["kibana_sample_data_flights", "kibana_sample_data_logs"]
```
It then uses this information to rewrite all table references to specified common database table. Rewriting table references in from clause should be sufficient for our current needs

### DIFF
--- a/quesma/quesma/config/config.go
+++ b/quesma/quesma/config/config.go
@@ -37,14 +37,15 @@ type QuesmaConfiguration struct {
 	//deprecated
 	ClickHouse RelationalDbConfiguration `koanf:"clickhouse"`
 	//deprecated
-	Hydrolix                   RelationalDbConfiguration     `koanf:"hydrolix"`
-	Elasticsearch              ElasticsearchConfiguration    `koanf:"elasticsearch"`
-	IndexConfig                map[string]IndexConfiguration `koanf:"indexes"`
-	Logging                    LoggingConfiguration          `koanf:"logging"`
-	PublicTcpPort              network.Port                  `koanf:"port"`
-	IngestStatistics           bool                          `koanf:"ingestStatistics"`
-	QuesmaInternalTelemetryUrl *Url                          `koanf:"internalTelemetryUrl"`
-	EnabledOptimizers          OptimizersConfiguration       `koanf:"optimizers"`
+	Hydrolix                      RelationalDbConfiguration             `koanf:"hydrolix"`
+	Elasticsearch                 ElasticsearchConfiguration            `koanf:"elasticsearch"`
+	IndexConfig                   map[string]IndexConfiguration         `koanf:"indexes"`
+	Logging                       LoggingConfiguration                  `koanf:"logging"`
+	PublicTcpPort                 network.Port                          `koanf:"port"`
+	IngestStatistics              bool                                  `koanf:"ingestStatistics"`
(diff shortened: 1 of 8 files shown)


quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -88,6 +99,11 @@ func Write(ctx context.Context, defaultIndex *string, bulk types.NDJSON, lm *cli
 			for _, document := range documents {
 				stats.GlobalStatistics.Process(cfg, indexName, document, clickhouse.NestedSeparator)
 			}
+			// if the index is mapped to specified database table in the configuration, use that table
+			sourceToDestinations := makeSourceToDestMappings(cfg.IndexSourceToInternalMappings)
+			if destIndex, ok := sourceToDestinations[indexName]; ok {
+				return lm.ProcessInsertQuery(ctx, destIndex, documents)
+			}
```
avelanarius, line 106: This is probably out-of-scope of this PR, but I wonder if we could unify "ingest transformers" in the future. For example now there's a `RewriteArrayOfObject` "ingest transformer" already here:

https://github.com/QuesmaOrg/quesma/blob/ebeaa89c994b62c9f0d7a402f7a8fe441f63ac3b/quesma/clickhouse/clickhouse.go#L499-L512

quesma/quesma/config/config.go
```diff
@@ -37,14 +37,15 @@ type QuesmaConfiguration struct {
...
-	EnabledOptimizers          OptimizersConfiguration       `koanf:"optimizers"`
+	Hydrolix                      RelationalDbConfiguration             `koanf:"hydrolix"`
+	Elasticsearch                 ElasticsearchConfiguration            `koanf:"elasticsearch"`
+	IndexConfig                   map[string]IndexConfiguration         `koanf:"indexes"`
+	Logging                       LoggingConfiguration                  `koanf:"logging"`
+	PublicTcpPort                 network.Port                          `koanf:"port"`
+	IngestStatistics              bool                                  `koanf:"ingestStatistics"`
+	QuesmaInternalTelemetryUrl    *Url                                  `koanf:"internalTelemetryUrl"`
+	EnabledOptimizers             OptimizersConfiguration               `koanf:"optimizers"`
+	IndexSourceToInternalMappings map[string]IndexMappingsConfiguration `koanf:"indexMappings"`
```
avelanarius, line 48: This is also out-of-scope of this PR, but I wonder if in the spirit of Quesma wanting to be programmable, maybe the (ingest) transformers could be more dynamic (e.g. their configuration not a static part of configuration, users being able to write their own transformers easily)

avelanarius (approved):
LGTM, my comments were just curiosity about our "North Star" (how it should look like in the future).

quesma/quesma/functionality/bulk/bulk.go
pdelewski: Probably we could do that, however this requires some changes in transformers API and as you mentioned this is out of scope of this particular PR

quesma/quesma/config/config.go
pdelewski: We can think about that

PR #533, opened 2024-07-15
CTEs as pointers

nablaone:
I think CTEs should be pointers just like other nodes in the AST.

@trzysiek  please take a look


trzysiek (approved):
sure, sorry for negligence

PR #534, opened 2024-07-15
Live view - show query time and transformations

nablaone:
In the "Live View" tab we display additional information for each executed SQL query: 
- duration time
- which schema transformation was applied
- which optimizations were performed 

Screenshot 2024-07-17 at 12 02 47


jakozaur (approved):
In future we may need it make it more compact. We may also want to include if we run them in parallel or sequential.

PR #536, opened 2024-07-15
Use schema registry information in CREATE TABLE

avelanarius:
When ingesting data into ClickHouse via Quesma, the first insert can automatically create a corresponding table in ClickHouse (if a user didn't do it beforehand).

That CREATE TABLE statement was previously based solely on the contents of the first inserted data (the first JSON). However, the information in that first JSON is often not sufficient to create a good ClickHouse schema. Prior to this change, the only way for the user to affect the CREATE TABLE was to perform it manually beforehand.

Extend the CREATE TABLE generation code to also take into account the information in schema the registry. Therefore the user can have some control over the CREATE TABLE statement via the "mapping" configuration in the config file. In practice this allows for:

1. Overriding the type of column (for example the initial JSON might have int values, but now the user can specify they should be stored as floats instead)
2. Adding columns that are missing in the initial JSON

One shortcoming of this solution is missing support for arrays. For example, `products.product_name: "text"` defined in "kibana_sample_data_ecommerce" is actually an array of text, but the configuration is not expressive enough to encode this fact yet. In such case the JSON type takes precedence over the user-specified configuration.

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -351,6 +351,13 @@ func (lm *LogManager) ProcessCreateTableQuery(ctx context.Context, query string,
 	return lm.sendCreateTableQuery(ctx, addOurFieldsToCreateTableQuery(query, config, table))
 }
 
+func findSchemaPointer(schemaRegistry schema.Registry, tableName string) *schema.Schema {
+	if foundSchema, found := schemaRegistry.FindSchema(schema.TableName(tableName)); found {
+		return &foundSchema
+	}
+	return nil
+}
+
 func buildCreateTableQueryNoOurFields(ctx context.Context, tableName string, jsonData types.JSON, tableConfig *ChTableConfig, cfg config.QuesmaConfiguration, schemaRegistry schema.Registry) (string, error) {
 
 	nameFormatter, err := registry.TableColumNameFormatterFor(tableName, cfg)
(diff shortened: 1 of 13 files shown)


quesma/clickhouse/parser.go
```diff
@@ -4,41 +4,75 @@ package clickhouse
...
+			result.WriteString(fmt.Sprintf("\"%s\" %s", columnFromSchema.ClickHouseColumnName, columnFromSchema.ClickHouseType))
+		} else {
+			result.WriteString(fmt.Sprintf("\"%s\" %s", columnFromJson.ClickHouseColumnName, columnFromJson.ClickHouseType))
 		}
 
+		delete(columnsFromSchema, schema.FieldName(columnFromJson.ClickHouseColumnName))
+	}
+
+	// There might be some columns from schema which were not present in the JSON
+	for _, column := range columnsFromSchema {
```
pdelewski, line 50: We use `first` in both loops, should not be `first` reseted before this loop?

quesma/clickhouse/clickhouse.go
```diff
@@ -349,18 +351,26 @@ func (lm *LogManager) ProcessCreateTableQuery(ctx context.Context, query string,
...
-func buildCreateTableQueryNoOurFields(ctx context.Context, tableName string, jsonData types.JSON, tableConfig *ChTableConfig, cfg config.QuesmaConfiguration) (string, error) {
+func buildCreateTableQueryNoOurFields(ctx context.Context, tableName string, jsonData types.JSON, tableConfig *ChTableConfig, cfg config.QuesmaConfiguration, schemaRegistry schema.Registry) (string, error) {
 
 	nameFormatter, err := registry.TableColumNameFormatterFor(tableName, cfg)
 	if err != nil {
 		return "", err
 	}
 
-	columns := FieldsMapToCreateTableString("", jsonData, 1, tableConfig, nameFormatter) + Indexes(jsonData)
+	var foundSchema *schema.Schema
```
pdelewski, line 361: I think that you can wrap this few lines as separate helper function that will return pointer to schema

quesma/clickhouse/parser.go
avelanarius: The `first` variable refers here to being first line in `result` and both of the loops write to `result` (so a write in the second loop is no longer first if there was a write in the first loop)

quesma/clickhouse/clickhouse.go
avelanarius: fixed in "review: extract finding schema pointer to findSchemaPointer" commit

quesma/clickhouse/parser.go
pdelewski: ok, thx for clarification

pdelewski (approved):

PR #539, opened 2024-07-17
10->1 SQL

trzysiek:


jakozaur:
We close it in favor of pancakes SQL generator.

PR #540, opened 2024-07-18
Optimizer - Rewrite query using a materialized view 

nablaone:
This PR adds an optimizer that rewrites queries and replaces the original table with a materialized view.  The rewrite rule is taken from the configuration. 

Suppose we've got the following table:
```
CREATE some_index (
  `date` DateTime64(3),
   `type` String,
   ... 
)
ENGINE=MergeTree()
ORDER BY (`date`)
```

We've discovered our dashboard generates a lot of queries with common expressions:
```
(type ILIKE '%foo%')
```

We may create a materialized view and query it instead of the original table.
Notice that the materialized view is updated on INSERT, it must be created before ingest.
```
-- a table that holds materialized view data 
CREATE table some_index_foo_dest (
  `date` DateTime64(3),
   `type` String,
   ... 
)

-- an index 

CREATE MATERIALIZED VIEW some_index_foo_mv to some_index_foo_dest AS
SELECT  * 
FROM some_index 
WHERE 
type ILIKE '%foo%'
```

Now we can configure the optimizer itself.
```
-- part of config.yaml
 some_index:
    enabled: true
    optimizers:
      materialized_view_replace:
        enabled: true
        properties:
          table: "some_index"
          condition: "(type ILIKE '%foo%')"
          view: "some_index_foo_mv"
```

So every query like
```
select count(*) from some_index where ((type ILIKE '%foo%') AND (`date` <= ..... AND `date` >= .....))
```
Will be rewritten to:
```
select count(*) from some_index where (TRUE AND (`date` <= ..... AND `date` >= .....))
```

Limitations:
- we can have only one rewrite rule per index/table
- a condition in the configuration must be the same as  a `string` representation of the  replaced expression 
- setup and configuration are  complex

### DIFF
--- a/quesma/clickhouse/quesma_communicator.go
+++ b/quesma/clickhouse/quesma_communicator.go
@@ -89,8 +89,8 @@ func (lm *LogManager) ProcessQuery(ctx context.Context, table *Table, query *mod
 
 var random = rand.New(rand.NewSource(time.Now().UnixNano()))
 
-const slowQueryThreshold = 2 * time.Second
-const slowQuerySampleRate = 1
+const slowQueryThreshold = 30 * time.Second
+const slowQuerySampleRate = 0.1
 
 func (lm *LogManager) shouldExplainQuery(elapsed time.Duration) bool {
 	return elapsed > slowQueryThreshold && random.Float64() < slowQuerySampleRate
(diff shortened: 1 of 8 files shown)


jakozaur (commented):
I worry if it's correct at all. For example I think it could replace expression inside `NOT` or `OR`. I would make it work only if given condition is on top level (or could be reached by AND from top level WHERE.

nablaone:
Yes. You're right. I've started examining arguments of  'AND' only and generalized too much.

jakozaur (approved):
The config cold be improved but let's do this in next PR.

PR #541, opened 2024-07-18
Fix a nil pointer dereference in GetSourceNames()

avelanarius:
Fix a nil pointer dereference in `GetSourceNames()`, which happened quite often for me while developing Quesma - I often start Quesma before Elastic is fully up, therefore `im.sources` would be nil (failed to load from Elastic) and `GetSourceNames()` would panic.

### DIFF
--- a/quesma/elasticsearch/index_management.go
+++ b/quesma/elasticsearch/index_management.go
@@ -53,6 +53,7 @@ func (im *indexManagement) GetSources() Sources {
 	if s := im.sources.Load(); s != nil {
 		return *s
 	} else {
+		logger.Warn().Msg("Indices are not yet loaded, returning empty sources.")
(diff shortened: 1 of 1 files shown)


quesma/elasticsearch/index_management.go
```diff
@@ -50,12 +50,16 @@ func (im *indexManagement) ReloadIndices() {
 }
 
 func (im *indexManagement) GetSources() Sources {
-	return *im.sources.Load()
+	if s := im.sources.Load(); s != nil {
```
pdelewski, line 53: I think we should log such situation

avelanarius: Added a warn message to that function.

pdelewski (approved):

PR #545, opened 2024-07-18
Add initial support for "PUT /:index" integrated with schema.Registry

avelanarius:
Add an initial support for the "PUT /:index" endpoint. This endpoint can be used to set a static mapping for an index and this is what the implementation focuses on.

After this change, a user can issue a "PUT /:index" before ingesting the data and the created table will take that information into the initial CREATE TABLE statement.

I have tested the solution with Flights and Ecommerce data and the schema information inferred from "PUT /:index" is sufficient to correctly ingest and query the data (removing all static configuration!).


quesma/elasticsearch/mappings.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/rs/zerolog/log"
+	"maps"
+	"quesma/elasticsearch/elasticsearch_field_types"
+	"quesma/schema"
+)
+
+func ParseMappings(namespace string, mappings map[string]interface{}) map[string]schema.Column {
+	result := make(map[string]schema.Column)
+
+	properties := mappings["properties"].(map[string]interface{})
```
pdelewski, line 15: I think it would be good to add sanity check to check whether "properties" key exists

quesma/schema/registry.go
```diff
@@ -157,3 +184,30 @@ func (s *schemaRegistry) populateSchemaFromTableDefinition(definitions map[strin
 	}
 	return found
 }
+
+// Used in tests:
+
+type StaticRegistry struct {
```
pdelewski, line 190: I think we can move it to separate file

quesma/quesma/router.go
```diff
@@ -254,9 +253,21 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
...
+	})
 
-	router.Register(routes.EQLSearch, and(method("GET", "POST"), matchedAgainstPattern(cfg)), eqlHandler)
+	router.Register(routes.IndexPath, and(method("PUT"), matchedAgainstPattern(cfg)), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
+		body, err := types.ExpectJSON(req.ParsedBody)
+		if err != nil {
+			return nil, err
+		}
+
+		mappings := body["mappings"].(map[string]interface{})
```
pdelewski, line 264: `PUT` is generally used for Index API and mapping is one of call. I think we should  whether it is mapping request

quesma/schema/registry.go
```diff
@@ -50,6 +53,25 @@ func (s *schemaRegistry) loadSchemas() (map[TableName]Schema, error) {
 	return schemas, nil
 }
 
+func (s *schemaRegistry) populateSchemaFromDynamicConfiguration(indexName string, fields map[FieldName]Field) {
+	if d, found := s.dynamicConfiguration[indexName]; found {
```
pdelewski, line 57: We can remove one level of indentation by reversing condition
```
if d, found := s.dynamicConfiguration[indexName]; !found {
  return
}
```

pdelewski (approved):

PR #549, opened 2024-07-19
Index mappings refinement

pdelewski:
This PR changes configuration and simplifies current implementation a lot.
It adds additional index property called `override`, example 

```
  kibana_sample_data_flights:
    enabled: true
    mappings:
      DestLocation: "geo_point"
      OriginLocation: "geo_point"
    override: big_kibana_common_table
```


avelanarius (approved):
Tested it with my Elastic Agent experimental setup, updated my config with the new `override` syntax, used the commits on this branch and it works correctly (both ingest and search of Docker Elastic Agent integration).

PR #551, opened 2024-07-19
Generate results in new combined query logic

trzysiek:
Issues:
* slow performance in Kibana. It's: 0,2sec Clickhouse query, 0,6sec response from Kibana. In other request it was 0,4sec and 0,9sec. Need to inspect what's the cause.
Screenshot 2024-07-19 at 18 10 00
Screenshot 2024-07-19 at 18 10 10


jakozaur:
Closed in favour of pancake logic.

PR #554, opened 2024-07-22
Execution plan

nablaone:
We are introducing `ExecutionPlan`. The idea is to keep everything related to processing a single elastic query in a single place. 

We currently keep only a list of queries to execute. 

There is a `QueryResultAdapter` added also. It will be used to implement an alternative way of processing some aggregations. That adapter will adapt alternative results set to the canonical one.

### DIFF
--- a/quesma/queryparser/query_translator.go
+++ b/quesma/queryparser/query_translator.go
@@ -4,6 +4,7 @@ package queryparser
 
 import (
 	"context"
+	"fmt"
 	"quesma/clickhouse"
 	"quesma/logger"
 	"quesma/model"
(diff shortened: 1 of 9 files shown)


quesma/queryparser/query_translator.go
```diff
@@ -169,6 +171,7 @@ func (cw *ClickhouseQueryTranslator) MakeAggregationPartOfResponse(queries []*mo
 	}
 	cw.postprocessPipelineAggregations(queries, ResultSets)
 	for i, query := range queries {
+		fmt.Printf("%d %s %v\n\n", i, model.AsString(query.SelectCommand), ResultSets[i])
```
jakozaur, line 174: Remove or comment.

jakozaur (approved):
Fine for now.

PR #558, opened 2024-07-23
Detect Kibana Fleet queries in `matchAgainstKibanaInternal`

avelanarius:
Before this change trying to access Kibana Fleet subpage in Kibana UI or trying to add a new integration would result in obscure Kibana errors. This was due to a fact that internal Kibana Fleet queries were handled by Quesma, instead of being routed to ElasticSearch.

Fix the problem by extending `matchAgainstKibanaInternal` function to be able to detect Kibana Fleet internal queries. After the fix you can now correctly access those previously failing Kibana Fleet subpages.

### DIFF
--- a/quesma/quesma/matchers.go
+++ b/quesma/quesma/matchers.go
@@ -133,6 +133,7 @@ func matchAgainstKibanaInternal() mux.RequestMatcher {
 
 		// 1. https://www.elastic.co/guide/en/security/current/alert-schema.html
 		// 2. migrationVersion
-		return !hasJsonKey("kibana.alert.", q) && !hasJsonKey("migrationVersion", q)
+		// 3., 4., 5. related to Kibana Fleet
+		return !hasJsonKey("kibana.alert.", q) && !hasJsonKey("migrationVersion", q) && !hasJsonKey("idleTimeoutExpiration", q) && !strings.Contains(req.Body, "fleet-message-signing-keys") && !strings.Contains(req.Body, "fleet-uninstall-tokens")
 	})
 }


avelanarius:
PR https://github.com/QuesmaOrg/quesma/pull/559 disables the `/_search` endpoint altogether, but I think it's still worth merging this PR in case we enable `/_search` back in the future.

quesma/quesma/matchers.go
```diff
@@ -133,6 +133,7 @@ func matchAgainstKibanaInternal() mux.RequestMatcher {
 
 		// 1. https://www.elastic.co/guide/en/security/current/alert-schema.html
 		// 2. migrationVersion
-		return !hasJsonKey("kibana.alert.", q) && !hasJsonKey("migrationVersion", q)
+		// 3., 4., 5. related to Kibana Fleet
+		return !hasJsonKey("kibana.alert.", q) && !hasJsonKey("migrationVersion", q) && !hasJsonKey("idleTimeoutExpiration", q) && !strings.Contains(req.Body, "fleet-message-signing-keys") && !strings.Contains(req.Body, "fleet-uninstall-tokens")
```
pdelewski, line 137: These keys could be extracted as consts

pdelewski (approved):

quesma/quesma/matchers.go
avelanarius: Those values are only used here once, I don't think we can give them good variable names (they are a bit of hack, we came up with them by looking at Kibana's traffic, trial-and-error approach).

So for now I'll leave it as-is, hardcoded here.

PR #559, opened 2024-07-23
Disable `/_search` (search over all indices)

avelanarius:
### Commit 1: "Disable /_search (search over all indices)"
Disable the `/_search` endpoint (an endpoint searching over`*` - all indices). This endpoint is used predominantly by internal Kibana queries and it's error-prone/time-consuming to detect those internal Kibana queries. Moreover, Quesma currently doesn't support stitching multiple tables, so a query over many indices wouldn't work well either way.

### Commit 2: "Replace Kibana Report with Kibana Discover in smoke test"
One of the smoke tests was a test of generating a Kibana Report. Unfortunately, Kibana internally generates this report by querying the `/_search` endpoint. Therefore this commit introduces an alternative way of making sure Kibana works correctly - it now uses an internal search endpoint which is used when one uses the Kibana Discover UI (code written based on observing requests it did).

### DIFF
--- a/quesma/quesma/mux/mux.go
+++ b/quesma/quesma/mux/mux.go
@@ -145,3 +145,13 @@ func (p *predicateAnd) Matches(req *Request) bool {
 func And(predicates ...RequestMatcher) RequestMatcher {
 	return &predicateAnd{predicates}
 }
+
+type predicateNever struct{}
+
+func (p *predicateNever) Matches(req *Request) bool {
+	return false
+}
+
+func Never() RequestMatcher {
+	return &predicateNever{}
+}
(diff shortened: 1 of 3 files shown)


avelanarius:
Converting to draft, the smoke-test is failing and I'll try to debug it.

avelanarius:
It turns out that Kiana report generation (report of inserted logs to `logs-generic` index) relies on the `/_search` endpoint and it just happened that `logs-generic` was selected by the "requires union of multiple tables [%s], not yet supported, picking just one" logic. After this PR, the global search is always handled by Elastic, but since it's empty the report is also empty...

avelanarius:
Since https://github.com/QuesmaOrg/quesma/pull/558 fixes Quesma for Kibana Fleet usecase, this PR is not that important now, I'll return to it later in the future.

avelanarius:
The PR is now ready to review, I dealt with problems with Kibana Report smoke test (by replacing it with a new Kibana Discover smoke test).

jakozaur (approved):

PR #562, opened 2024-07-23
Resolve override in `removeNotExistingTables`

avelanarius:
`removeNotExistingTables` is used in searches with wildcard index patterns to remove indexes which were enabled in a config but without a corresponding table in ClickHouse. However, the logic did not take into account the overrides. For example if index `A` was overridden to `B`, removeNotExistingTables would still search for ClickHouse table `A`, not `B` and wildcard search would return incomplete results.

A disadvantage of this solution is that we are doing Override resolution in yet another place, however I couldn't find a way to do that resolution earlier since later code still needs the original (pre-override) name.


avelanarius:
@pdelewski what's your opinion - should it be fixed this way?

pdelewski:
> @pdelewski what's your opinion - should it be fixed this way?

I'm ok with merging this for now

pdelewski (approved):

PR #563, opened 2024-07-23
Fix JsonToColumns bug generating empty Tuple type

avelanarius:
While testing Elastic Agent (Docker integration), I noticed that Quesma generated an invalid `CREATE TABLE` statement:

```
(...)
docker::cpu::core Tuple()
(...)
```

Having an empty Tuple is not allowed in ClickHouse. This type was inferred when receiving the following JSON field:

```
"core": {}
```

Fix the JsonToColumns function to ignore such a field instead of generating an empty Tuple column.

(Note that sometimes this field is non-empty, that's why Elastic Agent sent an empty one)

```
"core": {"pct":0.01963672211953173,"norm":{"pct":0.0012272951324707332}}
```

### DIFF
--- a/quesma/clickhouse/parser.go
+++ b/quesma/clickhouse/parser.go
@@ -73,18 +73,26 @@ func JsonToColumns(namespace string, m SchemaMap, indentLvl int, config *ChTable
 			nested := JsonToColumns(nameFormatter.Format(namespace, name), nestedValue, indentLvl, config, nameFormatter)
 			resultColumns = append(resultColumns, nested...)
 		} else {
-			var fType string
+			var fTypeString string
 			if value == nil { // HACK ALERT -> We're treating null values as strings for now, so that we don't completely discard documents with empty values
-				fType = "Nullable(String)"
+				fTypeString = "Nullable(String)"
 			} else {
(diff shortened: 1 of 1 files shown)


quesma/clickhouse/parser.go
```diff
@@ -73,20 +73,27 @@ func JsonToColumns(namespace string, m SchemaMap, indentLvl int, config *ChTable
...
+				fTypeString = "Nullable(String)"
 			} else {
-				fType = NewType(value).String()
-				if !strings.Contains(fType, "Array") && !strings.Contains(fType, "DateTime") {
-					fType = "Nullable(" + fType + ")"
+				fType := NewType(value)
+
+				// handle "field":{} case (Elastic Agent sends such JSON fields) by ignoring them
+				if multiValueType, ok := fType.(MultiValueType); ok && len(multiValueType.Cols) == 0 {
+					continue
```
pdelewski, line 84: It might be useful to generate log message in this case

avelanarius: Added a log message here and also rebased on current main.

pdelewski (approved):

PR #567, opened 2024-07-24
Fix handling of field names with `.` from incoming JSONs

jakozaur:
This is quick fix, currently we double create columns:
image
This removes original `host.name`.

I suggest next week we talk about properly addressing it.

### DIFF
--- a/quesma/clickhouse/parser.go
+++ b/quesma/clickhouse/parser.go
@@ -266,13 +266,6 @@ func RemoveTypeMismatchSchemaFields(m SchemaMap, t *Table) SchemaMap {
 		}
 	}
 	for fieldName, v := range m {
-		withoutDotsFieldName := strings.Replace(fieldName, ".", "::", -1)
-		if fieldName != withoutDotsFieldName {
-			m[withoutDotsFieldName] = v
-			delete(m, fieldName)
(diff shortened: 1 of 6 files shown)


quesma/clickhouse/parser.go
```diff
@@ -256,6 +266,13 @@ func RemoveTypeMismatchSchemaFields(m SchemaMap, t *Table) SchemaMap {
 		}
 	}
 	for fieldName, v := range m {
+		withoutDotsFieldName := strings.Replace(fieldName, ".", "::", -1)
+		if fieldName != withoutDotsFieldName {
+			m[withoutDotsFieldName] = v
+			delete(m, fieldName)
+			fieldName = withoutDotsFieldName
+		}
+
```
avelanarius, line 275: Isn't this portion of the code now unnecessary since you also modified `BuildInsertJson` (which calls this function, but after it already fixed all `.`s)?

jakozaur: I can refactor to one.

avelanarius:
I took the liberty to change the title from "Always create  column names." to more descriptive "Fix handling of field names with . from incoming JSONs"

avelanarius (approved):

PR #570, opened 2024-07-25
Adding ingest validation implementation

pdelewski:
This PR adds validation to data ingest.
If a json field has a type that is different from db column, it will be added to attributes.

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -423,8 +423,8 @@ func (lm *LogManager) CreateTableFromInsertQuery(ctx context.Context, name strin
 
 // This function takes an attributesMap and update it
 // with the fields that are not validated according to the inferred schema
-func updateAttributesMap(attrsMap map[string][]interface{}, notValidJson types.JSON) {
-	for k, v := range notValidJson {
+func addInvalidJsonFieldsToAttributes(attrsMap map[string][]interface{}, inValidJson types.JSON) {
+	for k, v := range inValidJson {
 		attrsMap[AttributesKeyColumn] = append(attrsMap[AttributesKeyColumn], k)
 		attrsMap[AttributesValueColumn] = append(attrsMap[AttributesValueColumn], v)
 	}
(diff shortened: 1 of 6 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -421,9 +421,18 @@ func (lm *LogManager) CreateTableFromInsertQuery(ctx context.Context, name strin
 	return nil
 }
 
+// This function takes an attributesMap and update it
+// with the fields that are not validated according to the inferred schema
+func addInvalidJsonFieldsToAttributes(attrsMap map[string][]interface{}, inValidJson types.JSON) {
```
avelanarius, line 426:
```suggestion
func addInvalidJsonFieldsToAttributes(attrsMap map[string][]interface{}, invalidJson types.JSON) {
```

here and in other places. The previous capitalization made it a bit confusing for me (it suggested "inValidJson = in valid JSON").

quesma/clickhouse/clickhouse.go
```diff
@@ -421,9 +421,18 @@ func (lm *LogManager) CreateTableFromInsertQuery(ctx context.Context, name strin
 	return nil
 }
 
+// This function takes an attributesMap and update it
+// with the fields that are not validated according to the inferred schema
```
avelanarius, line 425:
```suggestion
// This function takes an attributesMap and updates it
// with the fields that are not valid according to the inferred schema
```

quesma/clickhouse/clickhouse.go
```diff
@@ -473,6 +481,10 @@ func (lm *LogManager) BuildInsertJson(tableName string, data types.JSON, config
 	} else {
 		return "", fmt.Errorf("no attributes or others in config, but received non-schema fields: %s", mDiff)
 	}
+	// If there are some non-valid fields, we need to add them to the attributes map
```
avelanarius, line 484:
```suggestion
	// If there are some invalid fields, we need to add them to the attributes map
```

quesma/clickhouse/ingest_validator.go
```diff
@@ -0,0 +1,120 @@
...
+	incomingValueType := getTypeName(value)
+	if columnType == DateTimeType {
+		// TODO validate date format
+		// For now we store dates as strings
+		if incomingValueType != StringType {
+			deletedFields[fieldName] = value
+		}
+	} else if columnType != incomingValueType {
+		deletedFields[fieldName] = value
+	}
```
avelanarius, line 91: This should also allow inserting ints as floats (in one ingest data we examined, a user sometimes inserted "int" `0` to a float column)

avelanarius (commented):
Looks OK apart from the comments I left.

avelanarius (approved):

quesma/clickhouse/ingest_validator.go
pdelewski: Agree.

PR #574, opened 2024-07-25
Pancake optimization

jakozaur:
Way better way of generating SQL.


quesma/clickhouse/quesma_communicator.go
```diff
@@ -71,6 +72,18 @@ func (lm *LogManager) ProcessQuery(ctx context.Context, table *Table, query *mod
 			colName = col.ColumnName
```
pdelewski, line 72: IMO, this for loop should not be part of `ProcessQuery`. In ideal case this kind of transformation should be executed before calling `ProcessQuery`

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+package queryparser
+
+import (
+	"quesma/logger"
+	"quesma/model"
+)
+
+type pancakeAggregationTopLevel struct {
```
pdelewski, line 10: It seems that `pancakeAggregationTopLevel` is subtype of pancakeAggregationLevel, why do we have distinction between them, should not we have just one type?

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+	whereClause model.Expr
+}
+
+type pancakeAggregationLevel struct {
+	name            string
+	queryType       model.QueryType
+	selectedColumns []model.Expr
+
+	// only for bucket aggregations
+	children []*pancakeAggregationLevel
```
pdelewski, line 21: Maybe embedding `pancakeAggregationTopLevel` would be better and more clearly show intentions
```suggestion
	pancakeAggregationTopLevel
```

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+	orderBy  []model.OrderByExpr
+	limit    int // 0 if none, only for bucket aggregation
+	isKeyed  bool
+
+	metadata    model.JsonMap
+	whereClause model.Expr
```
pdelewski, line 27: As in above comment
```suggestion
```

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+type pancakeFillingMetricAggregation struct {
+	name            string
+	queryType       model.QueryType // it has to be metric aggregation
+	selectedColumns []model.Expr
+
+	metadata model.JsonMap
+}
+
+type pancakeLayerBucketAggregation struct {
+	name            string
```
pdelewski, line 39: `pancakeFillingMetricAggregation` contains all fields of pancakeFillingMetricAggregation, so the latter can be embedded

```suggestion
	pancakeFillingMetricAggregation
```

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+const PancakeTotalCountMetricName = "__quesma_total_count"
+
+// Not a real aggregation, but it is a pancake that has alternative JSON rendering
+type PancakeQueryType struct {
+	pancakeAggregation *pancakeAggregation
+}
+
+func (p PancakeQueryType) ReturnCount() *pancakeFillingMetricAggregation {
+
+	if len(p.pancakeAggregation.layers) > 0 {
```
pdelewski, line 73: Maybe reverse 
```suggestion
	if len(p.pancakeAggregation.layers) <= 0 {
	  return nil
	}
```

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+	"quesma/logger"
+	"quesma/model"
+)
+
+type pancakeAggregationTopLevel struct {
+	children    []*pancakeAggregationLevel
+	whereClause model.Expr
+}
+
+type pancakeAggregationLevel struct {
```
pdelewski, line 15: `pancakeAggregationLevel` contains everything from `pancakeFillingMetricAggregation`, embedding?

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+
+type pancakeFillingMetricAggregation struct {
+	name            string
+	queryType       model.QueryType // it has to be metric aggregation
+	selectedColumns []model.Expr
+
+	metadata model.JsonMap
+}
+
+type pancakeLayerBucketAggregation struct {
```
pdelewski, line 38: Would not just `pancakeBucketAggregation` be better name?

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+	children []*pancakeAggregationLevel
+	orderBy  []model.OrderByExpr
+	limit    int // 0 if none, only for bucket aggregation
+	isKeyed  bool
+
+	metadata    model.JsonMap
+	whereClause model.Expr
+}
+
+type pancakeFillingMetricAggregation struct {
```
pdelewski, line 30: Would not just pancakeMetricAggregation be better name?

quesma/queryparser/pancake_model.go
```diff
@@ -0,0 +1,105 @@
...
+	// only for bucket aggregations
+	orderBy []model.OrderByExpr
+	limit   int // 0 if none, only for bucket aggregation
+	isKeyed bool
+
+	metadata    model.JsonMap
+	whereClause model.Expr
+}
+
+type pancakeAggregationLayer struct {
```
pdelewski, line 52: Don't clearly understand difference between `Layer` and `Level`. Comment regarding model would be useful. I think that this model can be simplified.

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	indexName := rows[0].Index
+	buckets = append(buckets, model.QueryResultRow{Index: indexName})
+	subAggrs = append(subAggrs, []model.QueryResultRow{{Index: indexName}})
+	for _, cols := range rows[0].Cols {
+
+		if strings.HasPrefix(cols.ColName, name+"key") || strings.HasPrefix(cols.ColName, name+"count") {
+			buckets[0].Cols = append(buckets[0].Cols, cols)
+		} else {
+			subAggrs[0][0].Cols = append(subAggrs[0][0].Cols, cols)
```
pdelewski, line 48: is this `subAggrs[0][0]` guarantee to be valid?

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,180 @@
...
+	subAggrs := [][]model.QueryResultRow{}
+	if len(rows) == 0 {
+		return buckets, subAggrs
+	}
+	indexName := rows[0].Index
+	buckets = append(buckets, model.QueryResultRow{Index: indexName})
+	subAggrs = append(subAggrs, []model.QueryResultRow{{Index: indexName}})
+	for _, cols := range rows[0].Cols {
+
+		if strings.HasPrefix(cols.ColName, name+"key") || strings.HasPrefix(cols.ColName, name+"count") {
```
pdelewski, line 45: I think we could extract consts for these strings like `key`, `count` as they are used in many places to avoid typos

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,180 @@
...
+		metricName := "metric__"
+		for i := 0; i < layerId; i++ {
+			metricName = fmt.Sprintf("%s%s__", metricName, layers[i].nextBucketAggregation.name)
+		}
+		metricName = fmt.Sprintf("%s%s_col_", metricName, metric.name)
+		metricRows := p.selectMetricRows(metricName, rows)
+		result[metric.name] = metric.queryType.TranslateSqlResponseToJson(metricRows, 0) // TODO: fill level?
+	}
+
+	if layer.nextBucketAggregation != nil {
```
pdelewski, line 117:
```suggestion
	if layer.nextBucketAggregation == nil {
	    return result, nil
	}
```

quesma/queryparser/pancake_query_generation.go
```diff
@@ -0,0 +1,294 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"quesma/clickhouse"
+	"quesma/model"
+	"strconv"
+)
+
+type pancakeQueryGenerator struct {
```
pdelewski, line 13: Would not pancakeSqlQueryGenerator be more clear?

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,180 @@
...
+
+		if strings.HasPrefix(cols.ColName, name+"key") || strings.HasPrefix(cols.ColName, name+"count") {
+			buckets[0].Cols = append(buckets[0].Cols, cols)
+		} else {
+			subAggrs[0][0].Cols = append(subAggrs[0][0].Cols, cols)
+		}
+	}
+	// restRow
+	for _, row := range rows[1:] {
+		isNewBucket := false
```
pdelewski, line 53: Seems that `53-71` can be extracted as separate `isNewBucket` function

quesma/clickhouse/quesma_communicator.go
nablaone: We've now got two pipelines:
1. schema transformers
2. optimizers

Do you mean we should have the third "clean up/ compensation" pipeline?

pdelewski: I would suggest something simpler, just extract this functionality to separate function that will be called before `ProcessQuery`

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,160 @@
...
+// This can't be replaced by WHERE in generic case.
+func (p *pancakeJSONRenderer) potentiallyRemoveExtraBucket(layer *pancakeAggregationLayer, bucketName string, bucketRows []model.QueryResultRow, subAggrRows [][]model.QueryResultRow) ([]model.QueryResultRow, [][]model.QueryResultRow) {
+	// We are filter out null
+	if layer.nextBucketAggregation.filterOurEmptyKeyBucket {
+		nullRowToDelete := -1
+		bucketKeyName := bucketName + "key"
+	ROW:
+		for i, row := range bucketRows {
+			for _, col := range row.Cols {
+				if strings.HasPrefix(col.ColName, bucketKeyName) {
```
nablaone, line 85: We can find the column index before processing rows and use it instead of looping over cols

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -0,0 +1,160 @@
...
+
+	result := model.JsonMap{}
+	if layerIdx >= len(layers) {
+		return result, nil
+	}
+	layer := layers[layerIdx]
+	for _, metric := range layer.currentMetricAggregations {
+		metricName := ""
+		for _, prevLayer := range layers[:layerIdx] {
+			metricName = fmt.Sprintf("%s%s__", metricName, prevLayer.nextBucketAggregation.name)
```
nablaone, line 115: Maybe we should store these mangled names in the `pancakeAggregationLayer`.  Right now, the contract between the generator and renderer is fragile.

quesma/queryparser/pancake_model.go
nablaone: From my understanding `pancakeAggregationTopLevel` holds a forest, and `pancakeAggregationLevel` represents a single tree.

quesma/queryparser/pancake_model.go
nablaone: There is a process that transforms a tree of aggregation into a list of aggregation.

`pancakeAggregationLevel` is part of the input tree, `pancakeFillingMetricAggregation` is a part of the result.

I think we should keep them separate.

pdelewski: ok, but from content perspective they store the same set of fields, so maybe we can extract common part and embed it

quesma/queryparser/pancake_transformation.go
```diff
@@ -0,0 +1,143 @@
...
+// SPDX-License-Identifier: Elastic-2.0
+package queryparser
+
+import (
+	"fmt"
+	"quesma/model"
+	"sort"
+)
+
+type aggregationTree2Pancake struct {
```
trzysiek, line 11: some better name?

quesma/queryparser/pancake_transformation.go
```diff
@@ -0,0 +1,143 @@
...
+		orderBy:                 bucket.orderBy,
+		limit:                   bucket.limit,
+		isKeyed:                 bucket.isKeyed,
+		filterOurEmptyKeyBucket: bucket.filterOutEmptyKeyBucket,
+
+		metadata: bucket.metadata,
+	}, nil
+}
+
+func (a *aggregationTree2Pancake) bakeLayer(childAggregations []*pancakeAggregationLevel) (*pancakeAggregationLayer, *pancakeAggregationLevel, error) {
```
trzysiek, line 55: maybe named return values? Or a comment what they are

quesma/queryparser/query_translator.go
```diff
@@ -286,6 +299,63 @@ func (cw *ClickhouseQueryTranslator) makeTotalCount(queries []*model.Query, resu
...
+				}
+				total = &model.Total{
+					Value:    totalCount,
+					Relation: "eq",
+				}
+				return
+			}
+		}
+	}
+
```
trzysiek, line 358: Not too important, but this as well as 2 cases above should probably each be a separate function.

quesma/queryparser/pancake_json_rendering.go
jakozaur: I agree; it's easier to store or have common method.

quesma/queryparser/pancake_query_generation.go
```diff
@@ -0,0 +1,289 @@
...
+			// TODO: ...
+			for columnId, column := range bucketAggregation.selectedColumns {
+				aliasedName := fmt.Sprintf("aggr__%skey_%d", namePrefix, columnId)
+				// TODO: check for collisions
+				aliasedColumn := model.AliasedExpr{Expr: column, Alias: aliasedName}
+				selectedColumns = append(selectedColumns, aliasedColumn)
+				groupByColumns = append(groupByColumns, aliasedColumn)
+				addedGroupByAliases = append(addedGroupByAliases, p.newQuotedLiteral(aliasedName))
+			}
+
```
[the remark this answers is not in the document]
pdelewski: Yes, extensive comments would be very helpful

pdelewski:
As discussed we merge it as is and continue work on improvements

pdelewski (approved):

nablaone (approved):

jakozaur:
More cleanup work will happen in following PRs.

PR #575, opened 2024-07-26
Support sending bulk entries to both Elastic and ClickHouse

avelanarius:
_bulk requests can write data to many indexes. In particular, a bulk request could contain writes to both enabled and disabled indexes, so some parts of it should be sent to Elastic and some parts to ClickHouse.

Previously the code supported only either sending the entire request to Elastic or sending the inserts to ClickHouse. After this change, the code splits the bulk request into a Elastic and ClickHouse portions and can send it to both endpoints.

The response from Elastic is parsed and "multiplexed" into the final Quesma response (alongside results from ClickHouse). This required some code refactoring and now the code has a proper support for setting response codes/status per each inserted entry.

### DIFF
--- a/quesma/proxy/l4_proxy.go
+++ b/quesma/proxy/l4_proxy.go
@@ -106,17 +106,18 @@ func configureRouting() *http.ServeMux {
 			return
 		}
 
-		err := ndjson.BulkForEach(func(operation types.BulkOperation, document types.JSON) {
+		err := ndjson.BulkForEach(func(i int, operation types.BulkOperation, _, document types.JSON) error {
 
 			index := operation.GetIndex()
 			if index == "" {
 				logger.Error().Msg("No index in operation")
-				return
+				return nil
 			}
 
 			if !elasticsearch.IsInternalIndex(index) {
 				stats.GlobalStatistics.Process(configuration, index, document, clickhouse.NestedSeparator)
 			}
+			return nil
 		})
 
 		if err != nil {
--- a/quesma/proxy/l4_proxy.go
+++ b/quesma/proxy/l4_proxy.go
@@ -106,7 +106,7 @@ func configureRouting() *http.ServeMux {
 			return
 		}
 
-		err := ndjson.BulkForEach(func(i int, operation types.BulkOperation, _, document types.JSON) error {
+		err := ndjson.BulkForEach(func(entryNumber int, operation types.BulkOperation, _, document types.JSON) error {
 
 			index := operation.GetIndex()
 			if index == "" {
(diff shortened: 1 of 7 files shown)


quesma/proxy/l4_proxy.go
```diff
@@ -106,17 +106,18 @@ func configureRouting() *http.ServeMux {
 			return
 		}
 
-		err := ndjson.BulkForEach(func(operation types.BulkOperation, document types.JSON) {
+		err := ndjson.BulkForEach(func(i int, operation types.BulkOperation, _, document types.JSON) error {
```
pdelewski, line 109: I would suggest to change name of `i` parameter to be more meaningful

quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -3,110 +3,260 @@
...
 
-	for indexName, documents := range indicesWithDocumentsToInsert {
+	var elasticBulkResponse BulkResponse
+	err = json.Unmarshal(responseBody, &elasticBulkResponse)
+	if err != nil {
+		return err
+	}
+
+	// Copy Elastic's response entries to our response
+	for i, entry := range elasticBulkResponse.Items {
```
pdelewski, line 180: Is result of this copy used somewhere after `sendToElastic` function call?

quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -3,110 +3,260 @@
...
-			if err != nil {
-				logger.Error().Msgf("Error getting table column name formatter for index %s: %v", indexName, err)
-				return err
+			if err == nil {
+				tableMap, _ := lm.GetTableDefinitions()
+				transformer := registry.IngestTransformerFor(indexName, cfg, nil, tableMap)
+				err = lm.ProcessInsertQuery(ctx, indexName, inserts, transformer, nameFormatter)
+			}
+
+			for _, document := range documents {
```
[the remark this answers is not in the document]
avelanarius: Each document has a pointer to a corresponding response entry and this loop is filling up that response entry (`document.response.Create = ...` for example)

quesma/quesma/functionality/bulk/bulk.go
avelanarius: `.response` that I write to is a pointer into `results` and we return `results` in `Write` function.

Generally (maybe I should add some comment) the whole thing works like this: We receive a bulk request that writes to multiple indexes, we group those bulk entries by index. However, by doing this grouping we are losing the order - 1st bulk entry should correspond to 1st response entry, 2nd bulk entry to 2nd response entry and so on. To combat that, I create a `results` array and each `BulkRequestEntry` has a pointer to the correct entry in the resulting array.

pdelewski: I think comment would be useful

avelanarius: Added comment in the last commit.

quesma/proxy/l4_proxy.go
avelanarius: Change the parameter name from `i` to `entryNumber`

PR #576, opened 2024-07-29
Use quesma/quesma:latest in the kibana-sample-data example

pivovarit:


examples/run.sh
pivovarit, line 1: This file might not make sense anymore since the image is publicly available

PR #578, opened 2024-07-29
Add mechanism for validating the structure of discovered ClickHouse tables

pivovarit:
Created an internal mechanism for ClickHouse table structure validation - this will allow us to spot violations as early as possible. Since this is technology-specific, the validation mechanism is a part of the `clickhouse` package and is called when discovering existing tables.

For now, there's just one column name validation


quesma/clickhouse/table_discovery.go
```diff
@@ -275,6 +277,17 @@ func (sl *tableDiscovery) TableDefinitions() *TableMap {
 	return sl.tableDefinitions.Load()
 }
 
+func (sl *tableDiscovery) verify(tables map[string]discoveredTable) {
+	for _, table := range tables {
+		logger.Info().Msgf("verifying table %s", table.name)
```
nablaone, line 282: We should keep verification results in the table struct. We'll be able to show the "Problems" tab in the UI.

nablaone (approved):

quesma/clickhouse/table_discovery.go
pivovarit: Sounds like a great idea :)

PR #581, opened 2024-07-30
Update NOTICE.MD

pivovarit:
There's been a change in Quesma dependencies


pivovarit:
We need to reconfigure the job to use our machine account and then retry

PR #587, opened 2024-07-31
Rename `index_statistics` to `ingest_statistics`

avelanarius:
`index_statistics`/`IndexStatistics` is only used for collecting statistics about the ingest and it is NOT used for the query path. Rename the file, struct to better describe what types of statistics it actually collects.

Doing this change now to prevent future merge conflicts and in some preparation to potential future package reorganizations.


quesma/stats/ingest_statistics_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
```
avelanarius, line 1: This "new" file is just renamed `quesma/stats/index_statistics_test.go` with a couple `IndexStatistics` changed to `IngestStatistics`.

pdelewski (approved):

PR #589, opened 2024-07-31
Fix spurious use of indexResolver preventing idling

avelanarius:
We have noticed that Quesma still has problems with properly idling (not making queries to ClickHouse Cloud to allow it to idle).

The current idling logic relies on checking if table discovery was recently used. It turns out that idle Kibana very often does inserts to an internal `.kibana` index and to `/_monitoring/bulk` endpoint.

Even though those requests are routed to Elastic, we still checked if `.kibana` or `_monitoring` are names of enabled indices using indexResolver - this use bumped the LastAccessTime, preventing idling.

Fix the issue by avoiding a use of indexResolver if it's an endpoint (`_endpoint`) or if it's an internal index (starting with `.`).

### DIFF
--- a/quesma/feature/not_supported.go
+++ b/quesma/feature/not_supported.go
@@ -5,7 +5,9 @@ package feature
 import (
 	"context"
 	"fmt"
+	"quesma/elasticsearch"
 	"regexp"
+	"strings"
 )
 
 // https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html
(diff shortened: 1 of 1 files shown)


avelanarius:
@jakozaur I think that the idling mechanism based on the last access time of `IndexResolver` is too fragile, we should probably change it to something else in the future.

jakozaur:
@avelanarius 100% agree that current logic is really fragile, we should design something better.

jakozaur (approved):

PR #592, opened 2024-08-01
Ui report for ingest schema non-compliant fields

pdelewski:
Basic ui for schema non-compliant fields
image


nablaone:
There is a "global" ingest lock in the `process` function. We can replace `mutex` with channel.  We can gather or compute stats in the background (async).

pdelewski:
> There is a "global" ingest lock in the `process` function. We can replace `mutex` with channel. We can gather or compute stats in the background (async).

True, however that was out of the scope for this PR

nablaone (approved):

PR #596, opened 2024-08-02
Pancake column names

jakozaur:
Post merge pancake commits suggestions.

Centralize column name generation into one place.

### DIFF
--- a/quesma/queryparser/pancake_model.go
+++ b/quesma/queryparser/pancake_model.go
@@ -30,7 +30,7 @@ type pancakeAggregationLevel struct {
 
 type pancakeFillingMetricAggregation struct {
 	name            string          // as originally appeared in Query DSL
-	aliasName       string          // full name with path, e.g. metric__byCountry__byCity__population or aggr__byCountry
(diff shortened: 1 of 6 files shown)


quesma/queryparser/pancake_model.go
```diff
@@ -28,15 +29,17 @@ type pancakeAggregationLevel struct {
 }
 
 type pancakeFillingMetricAggregation struct {
-	name            string
+	name            string          // as originally appeared in Query DSL
+	aliasName       string          // full name with path, e.g. metric__byCountry__byCity__population or aggr__byCountry
```
pdelewski, line 33: I'm wondering whether `aliasName` is the best name for that, maybe `internalName`?

pdelewski (approved):

quesma/queryparser/pancake_model.go
nablaone: Or `mangledName`

jakozaur: Changed to `internalName`

PR #597, opened 2024-08-02
Pancake - update aggregation tests

trzysiek:
Before we had only a few tests for pancakes, they were also in a separate file `clients/ophelia_pancake.go`. Removed that file and added a pancake version for every aggregation test.

Also a few-line fix for terms with empty buckets. Before empty buckets case made rendering not work. A lot of tests failed because of that, so it was much easier for me to fix it here.


quesma/queryparser/aggregation_parser_test.go
```diff
@@ -723,6 +723,35 @@ func sortAggregations(aggregations []*model.Query) {
...
+	allTests = append(allTests, opensearch_visualize.PipelineAggregationTests...)
+	allTests = append(allTests, kibana_visualize.AggregationTests...)
+	allTests = append(allTests, clients.KunkkaTests...)
+	allTests = append(allTests, clients.OpheliaTests...)
+	return allTests
+}
+
+// TODO remove after pipeline aggregations are implemented
+func allAggregationTestsWithoutPipeline() []testdata.AggregationTestCase {
+	const lowerBoundTestNr = 80
```
jakozaur, line 743: Nit, but we don't need it. It is mere a very minor performance optimization in tests.

jakozaur (approved):

PR #605, opened 2024-08-05
Pancake `date histogram` 2

trzysiek:
Restores logic from old aggregation handling, so we add empty buckets in between if `min_doc_count == 0` in `(date_)histogram`

### DIFF
--- a/quesma/testdata/aggregation_requests.go
+++ b/quesma/testdata/aggregation_requests.go
@@ -5371,35 +5371,39 @@ var AggregationTests = []AggregationTestCase{
 				"total": 1
 			},
 			"aggregations": {
-				"2": {
+				"0": {
 					"buckets": [
 						{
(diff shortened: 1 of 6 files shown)


quesma/testdata/aggregation_requests.go
```diff
@@ -5371,35 +5371,39 @@ var AggregationTests = []AggregationTestCase{
...
-							"doc_count": 0,
-							"key": 9300.0
-						},
-						{
-							"doc_count": 0,
-							"key": 9400.0
-						},
-						{
-							"doc_count": 0,
-							"key": 9500.0
```
trzysiek, line 5394: Don't worry, this test was unfinished and had complete rubbish here, now it's actually a fully proper test.

jakozaur (approved):
Good step, but likely we need more consistent key for merging JSONs.

PR #611, opened 2024-08-05
Pancake `multi_terms` aggr

trzysiek:
old: There's one hack in the changes. Let's wait for #635 merge and then it might get fixed, not sure.
new: Yeah, some previous fixes made this PR very straightforward. Not much to comment on.


jakozaur (approved):
I still wish we will support multiple order by, but this is a major win.

PR #612, opened 2024-08-05
Fix parseTerm to resolve field name

avelanarius:
In most other `parse*` functions in `query_parser.go`, the incoming field names are first resolved with schema - this takes care of changing `.` to `::`. However, `parseTerm` was missing that call to `ResolveField` and just used the field name as-is (with `.`).

This caused the following query to not work properly:
```
GET index/_search
{
  "query": {
    "term": {
      "nested.clientip": "0.0.0.0/0"
    }
  }
}
```

Upon fixing `parseTerm`, `applyIpTransformations` had to be fixed - now it receives the "internal name" (`::`), not the original name (`.`) and has to resolve the field accordingly. Additionally, fields with `::` have to be quoted - not wrapped in un-quoted `model.LiteralExpr`.


pdelewski:
LGTM

pdelewski (approved):

PR #617, opened 2024-08-06
Remove PAT from `NOTICE.MD` generation job

mieciu:


pivovarit:
That won't work, changes pushed using this token won't trigger https://github.com/QuesmaOrg/quesma/blob/main/.github/workflows/generate-notice.yml

PR #624, opened 2024-08-07
Remove RemoveTypeMismatchSchemaFields (duplicate logic)

avelanarius:
While testing ingest of an `array` of objects with an `int` field, I noticed that this `int` field was not properly ingested. It turned out that the `RemoveTypeMismatchSchemaFields` logic removed that field - JSON numbers have a default `float` type, while the column in ClickHouse was an `int` column. The function had some logic to handle it, but it was missing from one code path.

However, `RemoveTypeMismatchSchemaFields` aims to do the same thing as the newly introduced ingest validation functionality (#570). Therefore this PR removes `RemoveTypeMismatchSchemaFields`, as the ingest validation already does its work.

I ported some tests of `RemoveTypeMismatchSchemaFields` into `ingest_validator.go`. During that I also made validation of arrays/slices more strict (validating each element, not just the first one).


jakozaur:
Thanks. @pdelewski likely would be the best person to review it.

quesma/clickhouse/ingest_validator.go
```diff
@@ -56,11 +56,25 @@ func getTypeName(v interface{}) string {
 		if len(elem) == 0 {
 			return arrayLiteral + "(unknown)"
 		} else {
-			return arrayLiteral + "(" + getTypeName(elem[0]) + ")"
+			innerTypeName := getTypeName(elem[0])
```
pdelewski, line 59: I think that we can extract this piece of code as separate function as use it in `interface{}` case as well

pdelewski:
Generally LGTM

quesma/clickhouse/ingest_validator.go
avelanarius: Is it OK if I also use reflection (`reflect.ValueOf`) in the case of arrays? Otherwise I don't see a good way to extract common logic (apart from extracting 2-3 lines to a separate function).

pdelewski: a ok. I'm not if that will have performance impact and if so, we can leave as it is

pdelewski (approved):

PR #626, opened 2024-08-08
Plugin removal

nablaone:
Plugins were premature.  We need to design a proper architecture first.


quesma/jsonprocessor/ingest.go
```diff
@@ -0,0 +1,43 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
```
avelanarius, line 1: The existing `quesma/jsonprocessor/json_processor.go` file is also used only for ingest, so maybe both of them should live in a single file.

However, I'm not sure if there should be a separate package`jsonprocessor (due to a risk of introducing circular dependencies later)

quesma/model/transformers.go
```diff
@@ -0,0 +1,17 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+package model
```
pdelewski, line 3: Not sure if that should be part of model, but we can revisit this later.

quesma/jsonprocessor/ingest.go
pdelewski: +1

pdelewski (approved):

jakozaur:
Let's talk about it during standup before merging.

jakozaur (approved):

PR #627, opened 2024-08-08
Fix responses of `_bulk` and `PUT /:index` (for Kafka Sink Connector)

avelanarius:
ElasticSearch Kafka Sink Connector uses the ElasticSearch Java Client v7.17.23 and it failed to deserialize some of the responses from our endpoints.

Firstly, a `_type` field has to be present in the `_bulk` response: https://github.com/elastic/elasticsearch/blob/61d76462eecaf09ada684d1b5d319b5ff6865a83/server/src/main/java/org/elasticsearch/action/DocWriteResponse.java#L116

Secondly, we were missing a response body for `PUT /:index` requests.

### DIFF
--- a/quesma/quesma/functionality/bulk/bulk.go
+++ b/quesma/quesma/functionality/bulk/bulk.go
@@ -54,6 +54,7 @@ type (
 		Result      string             `json:"result,omitempty"`
 		Status      int                `json:"status"`
 		Error       any                `json:"error,omitempty"`
+		Type        string             `json:"_type"` // ES 7.x Java Client requires this field
 	}
 	BulkShardsResponse struct {
 		Failed     int `json:"failed"`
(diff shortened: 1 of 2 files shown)


quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -54,6 +54,7 @@ type (
 		Result      string             `json:"result,omitempty"`
 		Status      int                `json:"status"`
 		Error       any                `json:"error,omitempty"`
+		Type        string             `json:"_type"`
```
mieciu, line 57: shouldn't we add `omitempty` here?

avelanarius: We always set it, so it's never empty. `Result`,`Error` is `omitempty` because we only fill one of them (`Result` for successes, `Error` for failures)

mieciu: I vaguely recall that it's been deprecated and no longer used in `8.x` (to be verified), but as long as it doesn't present any issues for any API clients - I'm be fine with that.

mieciu (approved):
LGTM

quesma/quesma/functionality/bulk/bulk.go
avelanarius: Added a comment here that some clients require this field.

PR #634, opened 2024-08-12
Pancake - `percentile_ranks` aggr

trzysiek:
Add `percentile_ranks` metric aggregation support in Pancakes.

### DIFF
--- a/quesma/model/metrics_aggregations/percentile_ranks.go
+++ b/quesma/model/metrics_aggregations/percentile_ranks.go
@@ -13,14 +13,14 @@ import (
 
 type PercentileRanks struct {
 	ctx         context.Context
-	columnNames []string
+	percentiles []string
 	// defines what response should look like
 	// https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-percentile-rank-aggregation.html#_keyed_response_5
 	Keyed bool
 }
 
-func NewPercentileRanks(ctx context.Context, columnNames []string, keyed bool) PercentileRanks {
-	return PercentileRanks{ctx: ctx, columnNames: columnNames, Keyed: keyed}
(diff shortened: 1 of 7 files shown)


quesma/model/metrics_aggregations/percentile_ranks.go
```diff
@@ -11,14 +11,15 @@ import (
 )
 
 type PercentileRanks struct {
-	ctx context.Context
+	ctx         context.Context
+	percentiles []string
```
jakozaur, line 15: I would add comment or even better name of that value. Not clear what it is.
```suggestion
	cutValues []string // count(if(field<$cutValue, 1, NULL))/count(*)*100
```

quesma/queryparser/aggregation_parser.go
```diff
@@ -228,8 +227,9 @@ func (b *aggrQueryBuilder) buildMetricsAggregation(metricsAggr metricsAggregatio
 			)
 			firstCountExp := model.NewFunction("count", ifExp)
```
jakozaur, line 228: While there, wouldn't be easier just to use `countIf` function?
https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators

It sounds simpler and likely ClickHouse could optimize it better.

quesma/model/metrics_aggregations/percentile_ranks.go
jakozaur, line 15: I would pick different name:
```
outvalues []string // DB, we count the percentage of records below the given value
```

quesma/queryparser/aggregation_parser.go
jakozaur: We may also do TODO, to replace and centralize second count(*).

jakozaur (commented):
We may want to improve two things while there.

jakozaur (approved):
Once you implement code fixes feel free to merge it.

PR #635, opened 2024-08-12
Pancake - `(date_)histogram` 1,5

trzysiek:
`(date_)histogram` wasn't working too well in pancakes, e.g. both new tests with simple 2x nested `(date_)histogram` used to fail.
Seems fixed now.

@jakozaur please look, generally problem is simple. We have such code:
`_, isColumnRef := orderBy.(model.ColumnRef)`
and then if true, we don't generate those `.part` columns. But e.g. in histogram we don't have a ColumnRef, but something very similar e.g. floor(column_ref_field/200)*200, and we don't want to group either.

LMK if you see a better way to do this than this `ColumnRefLike`

### DIFF
--- a/quesma/model/bucket_aggregations/histogram.go
+++ b/quesma/model/bucket_aggregations/histogram.go
@@ -4,6 +4,7 @@ package bucket_aggregations
 
 import (
 	"context"
+	"fmt"
 	"quesma/logger"
 	"quesma/model"
 	"quesma/util"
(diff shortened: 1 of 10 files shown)


quesma/model/bucket_aggregations/histogram.go
```diff
@@ -33,8 +33,8 @@ func (query Histogram) TranslateSqlResponseToJson(rows []model.QueryResultRow, l
 	var response []model.JsonMap
 	for _, row := range rows {
 		response = append(response, model.JsonMap{
-			"key":       row.Cols[level-1].Value,
-			"doc_count": row.Cols[level].Value,
+			"key":       row.Cols[len(row.Cols)-2].Value,
+			"doc_count": row.Cols[len(row.Cols)-1].Value,
```
trzysiek, line 37: let's maybe switch to this way, it's already like that in e.g. `histogram` and most other aggregations.
And start worrying only if we see it fail, which doesn't seem likely with amount of tests we have

jakozaur: Ideally would get right of level in many places, but probably after pancakes.

jakozaur (commented):
Ok, would also like to talk about it.

jakozaur (approved):
Please @trzysiek also look at it.

PR #637, opened 2024-08-13
Pancake - `sum_other_doc_count` in `terms` aggr

trzysiek:
Works only for pancakes.


quesma/model/bucket_aggregations/terms.go
```diff
@@ -56,3 +65,43 @@ func (query Terms) String() string {
...
+
+func (query Terms) docCount(row model.QueryResultRow) any {
+	return row.Cols[len(row.Cols)-1].Value
+}
+
+func (query Terms) key(row model.QueryResultRow) any {
+	return row.Cols[len(row.Cols)-2].Value
+}
+
+func (query Terms) parentCount(row model.QueryResultRow) any {
```
jakozaur, line 97: I don't like `parent_count`, took me a while before I understood it, why not `total_count`?

quesma/queryparser/pancake_sql_query_generation.go
```diff
@@ -105,6 +106,16 @@ func (p *pancakeSqlQueryGenerator) isPartOfGroupBy(column model.Expr, groupByCol
 func (p *pancakeSqlQueryGenerator) generateBucketSqlParts(bucketAggregation *pancakeModelBucketAggregation, groupByColumns []model.AliasedExpr, hasMoreBucketAggregations bool) (
 	addSelectColumns, addPartColumns, addGroupBys, addRankColumns []model.AliasedExpr, addRankWheres []model.Expr, addRankOrderBys []model.OrderByExpr, err error) {
 
+	if query_util.IsAnyKindOfTerms(bucketAggregation.queryType) {
+		parentCountColumn := model.WindowFunction{Name: "sum",
+			Args:        []model.Expr{model.NewFunction("count", model.NewLiteral("*"))},
```
jakozaur, line 111: Do we understnad how does it count null variables, do we need extra logic?

I would also prefer to make it after group by.

quesma/queryparser/pancake_model.go
```diff
@@ -65,3 +65,7 @@ func (p pancakeModelBucketAggregation) InternalNameForOrderBy(id int) string {
 func (p pancakeModelBucketAggregation) InternalNameForCount() string {
 	return fmt.Sprintf("%scount", p.internalName)
 }
+
```
jakozaur, line 68: Add comment.

PR #638, opened 2024-08-13
Removing others column

pdelewski:
This PR removes others as it is redundant to attributes (and maybe map in the future)

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -490,7 +490,7 @@ func (lm *LogManager) BuildInsertJson(tableName string, data types.JSON, inValid
 	}
 	var attrsMap map[string][]interface{}
 	if len(config.attributes) > 0 {
-		attrsMap, _ = BuildAttrsMapAndOthers(mDiff, config)
+		attrsMap, _ = BuildAttrsMap(mDiff, config)
 	} else {
 		return "", nil, fmt.Errorf("no attributes or others in config, but received non-schema fields: %s", mDiff)
 	}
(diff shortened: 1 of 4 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -498,11 +489,8 @@ func (lm *LogManager) BuildInsertJson(tableName string, data types.JSON, inValid
 		return js, nil, nil
 	}
 	var attrsMap map[string][]interface{}
-	var othersMap SchemaMap
 	if len(config.attributes) > 0 {
-		attrsMap, othersMap, _ = BuildAttrsMapAndOthers(mDiff, config)
-	} else if config.hasOthers {
-		othersMap = mDiff
+		attrsMap, _ = BuildAttrsMapAndOthers(mDiff, config)
```
avelanarius, line 493: Nit: you could rename `BuildAttrsMapAndOthers` to `BuildAttrsMap`

quesma/clickhouse/clickhouse_test.go
```diff
@@ -577,16 +565,16 @@ func TestJsonConvertingBoolToStringAttr(t *testing.T) {
...
 	assert.NoError(t, err)
-	assert.Equal(t, 0, len(others))
 	assert.Equal(t, 3, len(attrs))
 	for k := range attrs {
 		assert.Contains(t, k, "string")
 	}
 }
 
 func TestCreateTableString_1(t *testing.T) {
+	t.Skip("Others has been removed")
```
avelanarius, line 577: Maybe remove this test altogether then?

avelanarius (approved):
Looks OK, I suggested a couple more cleanups in my comments

quesma/clickhouse/clickhouse.go
pdelewski: Right, thx

quesma/clickhouse/clickhouse_test.go
pdelewski: You are right. I left it just to revisit this decision later

PR #642, opened 2024-08-14
Update README.md

ptbrzoska:
small detail


README.md
```diff
@@ -57,8 +57,7 @@ In order to run it, you need to:
 - run `docker-compose up` directly from `/examples/kibana-sample-data` directory
 
 If you want to build your own image, run `./bin/build-image.sh` from the root directory
-
-Handy one-liner:
+Handy one-liner for the above:
```
mieciu, line 60: That suggest that this one-liner will build the image. 
I'd suggest removing the line 59 about building the Docker image, it just irrelevant in this place and then it all start to make sense

PR #643, opened 2024-08-14
Pancakes column names

jakozaur:
Pancakes depend on unique names. The code needs to deal with edge cases well.

### DIFF
--- a/quesma/queryparser/pancake_transformer.go
+++ b/quesma/queryparser/pancake_transformer.go
@@ -16,12 +16,15 @@ type pancakeTransformer struct {
 	usedNames map[string][]string
 }
 
+func newPancakeTransformer() pancakeTransformer {
+	return pancakeTransformer{
+		usedNames: make(map[string][]string),
+	}
+}
+
 // Extremely rarely names may collide (e.g. aggregation named '1__2' and '1', '2').
 // This adds number suffix to make sure they are always unique
 func (a *pancakeTransformer) generateUniqueInternalName(origName string, aggrNames []string) string {
-	if a.usedNames == nil {
-		a.usedNames = make(map[string][]string)
-	}
 	proposedName := origName
 	for counter := 2; true; counter += 1 {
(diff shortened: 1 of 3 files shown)


quesma/queryparser/pancake_transformer.go
```diff
@@ -6,12 +6,43 @@ import (
...
 type pancakeTransformer struct {
+	usedNames map[string][]string
+}
+
+// Extremely rarely names may collide (e.g. aggregation named '1__2' and '1', '2').
+// This adds number suffix to make sure they are always unique
+func (a *pancakeTransformer) generateUniqueInternalName(origName string, aggrNames []string) string {
+	if a.usedNames == nil {
+		a.usedNames = make(map[string][]string)
+	}
```
trzysiek, line 24:
```suggestion
	usedNames map[string][]string
}

func newPancakeTransformer() pancakeTransformer {
    return pancakeTransformer{make(map[string][]string}
}

// Extremely rarely names may collide (e.g. aggregation named '1__2' and '1', '2').
// This adds number suffix to make sure they are always unique
func (a *pancakeTransformer) generateUniqueInternalName(origName string, aggrNames []string) string {
```

trzysiek, line 24:
```suggestion
func newPancakeTransformer() pancakeTransformer {
    return pancakeTransformer{make(map[string][]string}
}

// Extremely rarely names may collide (e.g. aggregation named '1__2' and '1', '2').
// This adds number suffix to make sure they are always unique
func (a *pancakeTransformer) generateUniqueInternalName(origName string, aggrNames []string) string {
```

trzysiek: Nitpick, but I'd change it to this, and use this constructor. This way we don't need to worry about concurrency in this `nil` check.

trzysiek (approved):
easy and clean implementation, I like it!

quesma/queryparser/pancake_transformer.go
jakozaur: Ok, will do.

PR #644, opened 2024-08-14
Add "PUT /:index/_mapping" endpoint

avelanarius:
Add a new `PUT /:index/_mapping` endpoint for static mappings. The implementation is similar to the existing `PUT /:index` endpoint, which can also contain a (static) mapping.

Additionally add a previously missing 'double' case to parseElasticType and add a warning message if the parsing failed.


avelanarius:
(I'm splitting my work to 2 PRs, next one will contain the more complicated `GET /:index/_mapping`)

pdelewski (approved):

PR #647, opened 2024-08-14
Accum agg pancake function

jakozaur:
Elastic got approximate cardinality, we should do the same.

Also implement all currently used functions.

### DIFF
--- a/quesma/queryparser/pancake_sql_query_generation.go
+++ b/quesma/queryparser/pancake_sql_query_generation.go
@@ -50,12 +50,12 @@ func (p *pancakeSqlQueryGenerator) generateAccumAggrFunctions(origExpr model.Exp
 	switch origFunc := origExpr.(type) {
 	case model.FunctionExpr:
 		switch origFunc.Name {
-		case "sumOrNull", "minOrNull", "maxOrNull":
+		case "sum", "sumOrNull", "min", "minOrNull", "max", "maxOrNull":
 			return origExpr, origFunc.Name, nil
-		case "avgOrNull":
-			return model.NewFunction("avgState", origFunc.Args...), "avgMerge", nil
-		case "count":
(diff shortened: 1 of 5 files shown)


mieciu (approved):
neat! 👍

quesma/queryparser/pancake_sql_query_generation.go
```diff
@@ -45,17 +45,18 @@ func (p *pancakeSqlQueryGenerator) generatePartitionBy(groupByColumns []model.Al
...
 			return origExpr, origFunc.Name, nil
-		case "avgOrNull":
-			return model.NewFunction("avgState", origFunc.Args...), "avgMerge", nil
-		case "count":
-			return model.NewFunction("count", origFunc.Args...), "sum", nil
+		case "count", "countIf":
+			return model.NewFunction(origFunc.Name, origFunc.Args...), "sum", nil
+		case "avg", "avgOrNull", "varPop", "varSamp", "stddevPop", "stddevSamp", "uniq":
+			// TODO: I debate whether make that default
+			return model.NewFunction(origFunc.Name+"State", origFunc.Args...), origFunc.Name + "Merge", nil
```
nablaone, line 59: This is Clickhouse specific. I'm not sure if the other databases have the same concept. 

It may require some comment or link to the documentation.

https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators

nablaone (commented):
Looks good. 

Please take a look at comment. These combinators are clever idea, but not so obvious.

quesma/queryparser/pancake_sql_query_generation.go
jakozaur: True.

Some databases like Snowflake got equivalent or at least for some operations (e.g. distinct count with HyperLogLog). Some can be emulated (e.g. average), but this is not an universal feature.

PR #648, opened 2024-08-14
AST - add JOIN expression

nablaone:


mieciu (approved):
Code-wise LGTM. 
I'm a little partial to adding it before having the application, but I guess upcoming PRs will be an easier read 😉

PR #652, opened 2024-08-19
Fixing issue with first insert with new field and losing data

pdelewski:


quesma/clickhouse/alter_table_test.go
```diff
@@ -0,0 +1,54 @@
...
+		engine:               "MergeTree",
+		orderBy:              "(timestamp)",
+		partitionBy:          "",
+		primaryKey:           "",
+		ttl:                  "",
+		attributes: []Attribute{
+			NewDefaultStringAttribute(),
+		},
+		castUnsupportedAttrValueTypesToString: true,
+		preferCastingToOthers:                 true,
```
avelanarius, line 26: Unrelated to this PR, but since `others` were removed, `preferCastingToOthers` should also be removed?

avelanarius (approved):

quesma/clickhouse/alter_table_test.go
pdelewski: Yes, true. We can do this cleanup as followup

PR #653, opened 2024-08-19
Fix pancake order by

jakozaur:
I simplified the order by implementing it in the first revision of the pancakes. It ignored all but the first order and added its own for dense_rank.

Now:
- We generate all order by's. We need that for correctness.
- We no longer generate order columns if they were part of the key (e.g. `date_histogram`).
- We deduplicate sort.
- We also replaced `Exprs []model.Expr` with `Expr model.Expr` previously we never used it and this was a trap.

This make SQL shorter and more performant. It turn out I had to implement that optimization to make queries correct.


nablaone (approved):
LGTM

PR #658, opened 2024-08-21
Add "get mapping" endpoints (GET /:index, /:index/_mapping)

avelanarius:
This commit adds support for `GET /:index` and `GET /:index/_mapping` endpoints. The primary use of those endpoints is to get the (static) mapping of an index. Those endpoints are actively used by some Elastic clients, such as ElasticSearch Kafka Sink Connector.

The implementation generates the mapping based on the schema found in the schema registry.

`hierarchical_schema.go` was introduced since the schema registry stores the schema in a "flat" format (e.g. `products.product_id`, `products.base_price`), but the mapping we need to return should be nested. `SchemaToHierarchicalSchema` in that file transforms the flat schema into a tree (for example `products` is a node with `product_id`, `base_price` children), which is more convenient to use.

The added tests in `mappings_test.go` show that a round-trip should be possible: sending a mapping to Quesma, it getting parsed and stored as a `Schema` struct and generating back a mapping based on that `Schema` struct representation. (It's not 100% equal, since we change numeric types to wider types, for example floats to doubles.)

### DIFF
--- a/quesma/quesma/router.go
+++ b/quesma/quesma/router.go
@@ -49,7 +49,7 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
 		}
 
 		results, err := bulk.Write(ctx, nil, body, lm, cfg, phoneHomeAgent)
-		return bulkInsertResult(results, err), nil
+		return bulkInsertResult(results, err)
 	})
 
 	router.Register(routes.IndexRefreshPath, and(method("POST"), matchedExact(cfg)), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
@@ -64,7 +64,11 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
 		}
 
 		err = doc.Write(ctx, req.Params["index"], body, lm, cfg)
-		return indexDocResult(req.Params["index"], http.StatusOK), err
+		if err != nil {
+			return nil, err
+		}
+
+		return indexDocResult(req.Params["index"], http.StatusOK)
 	})
 
 	router.Register(routes.IndexBulkPath, and(method("POST", "PUT"), matchedExact(cfg)), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
@@ -76,7 +80,7 @@ func configureRouter(cfg config.QuesmaConfiguration, sr schema.Registry, lm *cli
 		}
 
 		results, err := bulk.Write(ctx, &index, body, lm, cfg, phoneHomeAgent)
-		return bulkInsertResult(results, err), nil
+		return bulkInsertResult(results, err)
 	})
 
 	router.Register(routes.ResolveIndexPath, method("GET"), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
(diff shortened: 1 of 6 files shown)


quesma/schema/hierarchical_schema.go
```diff
@@ -0,0 +1,47 @@
...
+			newNode := SchemaTreeNode{Name: component}
+			currentNode.Children = append(currentNode.Children, &newNode)
+			currentNode = &newNode
+		}
+	}
+
+	currentNode.Children = append(currentNode.Children, &SchemaTreeNode{Name: leafComponent, Field: field})
+}
+
+func SchemaToHierarchicalSchema(s *Schema) *SchemaTreeNode {
```
mieciu, line 39: I think it's worth putting a comment over this method about what it does (the "unflattening") with an example. 
PR description is clear, but here it'll be more accessible in the future :)

quesma/quesma/router.go
```diff
@@ -425,6 +453,26 @@ func putIndexResult(index string) *mux.Result {
...
 
+func getIndexMappingResult(index string, mappings map[string]any) *mux.Result {
+	result := map[string]any{
+		index: map[string]any{
+			"mappings": mappings,
+		},
+	}
+	serialized, err := json.Marshal(result)
+	if err != nil {
+		panic(err)
```
mieciu, line 464: let's avoid using panic's at all cost. 
Throwing 500 will be more appropriate and will keep Quesma going

jakozaur: +1

quesma/schema/hierarchical_schema.go
```diff
@@ -0,0 +1,47 @@
...
+	components = components[:len(components)-1]
+
+	currentNode := root
+	for _, component := range components {
+		found := false
+		for _, child := range currentNode.Children {
+			if child.Name == component {
+				currentNode = child
+				found = true
+				break
```
jakozaur, line 26: I'm debating whether removing `found`, namping COMPONENT_LOOP and using continue COMPONENT_LOOP would make more sense here.

jakozaur (approved):
Someone else than me should also review it.

mieciu (changes requested):
Just a small note on `panic` and comment worth putting, other than that - LGTM

quesma/quesma/router.go
avelanarius: Fixed in https://github.com/QuesmaOrg/quesma/pull/658/commits/97bfee9cc80fb958042feb871a4bc932844af234

(also fixed all other `panic`s in that file - all shared the same pattern)

quesma/schema/hierarchical_schema.go
avelanarius: Added comment in https://github.com/QuesmaOrg/quesma/pull/658/commits/5cc46998a0aecbb14e912433eb21d81593c8b1e8

quesma/schema/hierarchical_schema.go
avelanarius: It would look like this:
```go
COMPONENT_LOOP:
	for _, component := range components {
		for _, child := range currentNode.Children {
			if child.Name == component {
				currentNode = child
				continue COMPONENT_LOOP
			}
		}

		newNode := SchemaTreeNode{Name: component}
		currentNode.Children = append(currentNode.Children, &newNode)
		currentNode = &newNode
	}
```

In that approach I don't like that at first sight those three lines:
```
		newNode := SchemaTreeNode{Name: component}
		currentNode.Children = append(currentNode.Children, &newNode)
		currentNode = &newNode
```
seem like they are executed in all iterations unconditionally, only at longer inspection a reader figures it out (that it's skipped with `continue`).

Therefore, I'm leaving this as-is for now.

mieciu (approved):
lgtm 👍

PR #659, opened 2024-08-21
Update README.md

ptbrzoska:
1. updated diagram
2. updated getting started commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ The setup contains following demo datasets:
 ### Running the example
 In order to run it, you need to:
 - clone this repository: `git clone https://github.com/QuesmaOrg/quesma.git`
-- run `docker-compose up` directly from `/examples/kibana-sample-data` directory
+- run `docker-compose up` directly from `examples/kibana-sample-data` directory
 
 Handy one-liner for above commands:
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,12 +53,12 @@ The setup contains following demo datasets:
 
 ### Running the example
 In order to run it, you need to:
-- clone this repository: `git clone git@github.com:QuesmaOrg/quesma.git`
+- clone this repository: `git clone https://github.com/QuesmaOrg/quesma.git`
 - run `docker-compose up` directly from `/examples/kibana-sample-data` directory
```
mieciu, line 57:
```suggestion
- run `docker-compose up` directly from `examples/kibana-sample-data` directory
```

mieciu (approved):
one nit spotted (unrelated to your changes, but worth adding if we're already working on this), other than that LGTM

README.md
avelanarius, line 57:
```suggestion
- `cd quesma/examples/kibana-sample-data`
- run `docker-compose up`
```

or maybe like this? This way it's easily copy-paste'able

ptbrzoska: Let's leave it as it was and have ppl to decide how they want to do it. If someone needs a command - it is provided below.

PR #660, opened 2024-08-21
Many pancakes out of one query

jakozaur:
Pancakes have quite a few limitations, and only one pancake can be created.

This allows more pancakes to be created per aggregation which brings it closer to feature completion.

### DIFF
--- a/quesma/queryparser/pancake_transformer.go
+++ b/quesma/queryparser/pancake_transformer.go
@@ -108,6 +108,13 @@ func (a *pancakeTransformer) createLayer(previousAggrNames []string, childAggreg
 		currentMetricAggregations: make([]*pancakeModelMetricAggregation, 0),
 	}
 
+	// we need sort aggregation to generate consistent results, otherwise:
+	// - tests are flaky
+	// - database might not use it query cache
+	sort.Slice(childAggregations, func(i, j int) bool {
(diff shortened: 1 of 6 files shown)


quesma/queryparser/pancake_transformer.go
```diff
@@ -91,100 +91,128 @@ func (a *pancakeTransformer) bucketAggregationToLayer(previousAggrNames []string
...
 }
 
-func (a *pancakeTransformer) aggregationTreeToPancake(topLevel pancakeAggregationTree) (pancakeResult *pancakeModel, err error) {
+func (a *pancakeTransformer) aggregationTreeToPancakes(topLevel pancakeAggregationTree) (pancakeResults []*pancakeModel, err error) {
 	if topLevel.children == nil || len(topLevel.children) == 0 {
 		return nil, fmt.Errorf("no top level aggregations found")
 	}
 
-	var nextBucketAggregation *pancakeAggregationTreeNode
+	resultLayers, err := a.aggregationChildrenToLayers([]string{}, topLevel.children)
```
nablaone, line 196: resultLayers can be nil. It's not handled.

nablaone: It's OK.  nil slice is empty.

nablaone (approved):

PR #663, opened 2024-08-22
Alter column heuristic

pdelewski:
This PR: 
- enables altering column mechanism as default
- introduce simple heuristic (first N columns is added by default, next columns are added based on simple statistic calculation)
- opens a way for changing/extending current heuristic

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -485,13 +485,8 @@ func generateNonSchemaFieldsString(attrsMap map[string][]interface{}) (string, e
 func (lm *LogManager) alterColumns(table *Table) bool {
 	// Above this number of columns we will use heuristic
 	// to decide if we should add new columns
-	/*
-		const maxColumns = 100
-		if len(table.Cols) < maxColumns {
-			return true
-		}
-	*/
-	return false
+	const maxColumns = 100
+	return len(table.Cols) < maxColumns
 }
 
 func (lm *LogManager) BuildIngestSQLStatements(tableName string, data types.JSON, inValidJson types.JSON,
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -32,7 +32,7 @@ const (
 	timestampFieldName = "@timestamp" // it's always DateTime64 for now, don't want to waste time changing that, we don't seem to use that anyway
 	// Above this number of columns we will use heuristic
 	// to decide if we should add new columns
-	maxColumns = 100
+	alwaysAddColumnLimit = 100
 )
 
 type (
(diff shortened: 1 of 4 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -482,8 +496,38 @@ func generateNonSchemaFieldsString(attrsMap map[string][]interface{}) (string, e
 	return nonSchemaStr, nil
 }
 
+// This function implements heuristic for deciding if we should add new columns
+func (lm *LogManager) shouldAlterColumns(table *Table, attrsMap map[string][]interface{}) (bool, []int) {
```
jakozaur, line 500: Ok, I think the below model should work:
1. Below 100, always add.
2. Till 1000 use heuristic.
3. Don't add above that number.

quesma/clickhouse/clickhouse.go
```diff
@@ -482,8 +496,38 @@ func generateNonSchemaFieldsString(attrsMap map[string][]interface{}) (string, e
...
+	attrKeys := getAttributesByArrayName(AttributesKeyColumn, attrsMap)
+	alterColumnIndexes := make([]int, 0)
+	if len(table.Cols) < maxColumns {
```
jakozaur, line 503: This should be configurable, but feel free to postpone after this PR.

Though let's rename this constant `alwaysAddColumnLimit`.

quesma/clickhouse/clickhouse.go
```diff
@@ -482,8 +496,38 @@ func generateNonSchemaFieldsString(attrsMap map[string][]interface{}) (string, e
...
+		// therefore we need to add all attrKeys indexes to alterColumnIndexes
+		for i := 0; i < len(attrKeys); i++ {
+			alterColumnIndexes = append(alterColumnIndexes, i)
+		}
+		return true, alterColumnIndexes
+	}
+	if lm.ingestFieldStatistics == nil {
+		lm.ingestFieldStatistics = make(IngestFieldStatistics)
+	}
+	const percent50 = 0.5
```
jakozaur, line 514: Very strict, would almost never work. E.g. if after one day, new column always appear, it will be rejected.

In first iteration, we can simplify that will do it after X counts. Where X is 100 or 1000.

quesma/clickhouse/clickhouse.go
```diff
@@ -30,18 +30,31 @@ import (
...
+	// to decide if we should add new columns
+	maxColumns = 100
+)
+
+type (
+	IngestFieldBucketKey struct {
+		indexName string
+		field     string
+	}
+	IngestFieldStatistics map[IngestFieldBucketKey]int64
```
jakozaur, line 43: Not sure if it's thread safe, would take the lock.

pdelewski: Yes. Good point

quesma/clickhouse/clickhouse.go
pdelewski: ok

jakozaur:
Great job!

quesma/clickhouse/clickhouse.go
pdelewski: As discussed added upper limit for now. We can simplify this in next PR

jakozaur (approved):

PR #664, opened 2024-08-22
Introducing new configuration syntax

mieciu:
I ended up doing very simple mapping from new config to the existing one. 

After few experiments, this turned out to be the only reasonable way to go as the codebase today doesn't play well with the target configuration syntax. Therefore, injecting constructs from the new config and propagating them over existing modules ended up bringing a lot mess to many different places.


quesma/quesma/config/config_v2.go
```diff
@@ -0,0 +1,169 @@
...
+	conf.InstallationId = c.InstallationId
+	conf.LicenseKey = c.LicenseKey
+	conf.IngestStatistics = c.IngestStatistics
+	conf.Connectors = make(map[string]RelationalDbConfiguration)
+	relDBConn, connType, err := c.getRelationalDBConf()
+	if err != nil {
+		errAcc = multierror.Append(errAcc, err)
+	} else {
+		relDBConn.ConnectorType = connType
+		if connType == "hydrolix" {
```
pdelewski, line 88: Would suggest to extract these strings as consts or enums

quesma/quesma/config/config_v2.go
```diff
@@ -0,0 +1,177 @@
...
+func (c *QuesmaNewConfiguration) getRelationalDBConf() (*RelationalDbConfiguration, string, error) {
+	if backendConn, typ := c.getRelationalDBBackendConnector(); backendConn != nil {
+		return &backendConn.Config, typ, nil
+	}
+	return nil, "", errors.New("exactly one backend connector of type `clickhouse`, `clickhouse-os` or `hydrolix` must be configured")
+}
+
+func (c *QuesmaNewConfiguration) getProcessorConfig() (*Processor, error) {
+	if len(c.Processors) == 1 {
+		if c.Processors[0].Type == "quesma-v1-processor" {
```
pdelewski, line 170: nitpick: I think `quesma-v1-processor` could be enum

quesma/quesma/config/config_v2.go
```diff
@@ -0,0 +1,177 @@
...
+	conf.LicenseKey = c.LicenseKey
+	conf.IngestStatistics = c.IngestStatistics
+	conf.Connectors = make(map[string]RelationalDbConfiguration)
+	relDBConn, connType, err := c.getRelationalDBConf()
+	if err != nil {
+		errAcc = multierror.Append(errAcc, err)
+	} else {
+		relDBConn.ConnectorType = connType
+		if connType == HydrolixBackendConnectorName {
+			conf.Connectors["injected-hydrolix-connector"] = *relDBConn
```
pdelewski, line 97: Maybe `injected-hydrolix-connector` and `injected-clickhouse-connector` extract as const or enum

ci/quesma/config/ci-config.yaml
```diff
@@ -1,15 +1,33 @@
...
+      url: "http://elasticsearch:9200"
+  - name: my-clickhouse-cloud-data-source
+    type: clickhouse-os
+    config:
+      url: "clickhouse://clickhouse:9000"
+processors:
+  - name: elasticsearch-to-clickhouse-processor
+    type: quesma-v1-processor
+    config:
+      mode: "dual-write-query-clickhouse"
```
pdelewski, line 20: I'm wondering whether `dual-write-query-clickhouse` effectively mean/implies that `enableElasticsearchIngest` should be true.

quesma/quesma/config/config_v2.go
mieciu: I didn't wanted to do that because that's just the "working name" which isn't used/referred anywhere. This will also be gone once we remove the old config object :)

ci/quesma/config/ci-config.yaml
mieciu: WDYT about introducing `disableElasticsearchIngest` flag then?

mieciu: (yea I agree the implication is the opposite to what I'd expect)

pdelewski (approved):

ci/quesma/config/ci-config.yaml
mieciu: 👍 so let it stay as is for now to maintain the default behavior and we'll clean this up as we revisit modes.

PR #665, opened 2024-08-22
Pancake filter implementation

jakozaur:
Now, we can use filters in pancakes:
1. It fixes many panels in flight sample dashboards.
2. Feature parity and even exceed original implementation.

While implementing that, I needed to fix some more issues:
- the `doc_count` was ignored in tests, fixed it while also addressing underlying issue
- our query simplification got bugs and fixed that too

The code is a bit more complex, but I prefer smaller ones. After `filters`, `range`, and `dataRange`, I will do the final refactor, though I still need to think through a few cases.

### DIFF
--- a/quesma/queryparser/pancake_json_rendering.go
+++ b/quesma/queryparser/pancake_json_rendering.go
@@ -119,28 +119,24 @@ func (p *pancakeJSONRenderer) layerToJSON(layerIdx int, layers []*pancakeModelLa
 	}
 
 	if layer.nextBucketAggregation != nil {
-		// sampler is special
-		if _, isSampler := layer.nextBucketAggregation.queryType.(bucket_aggregations.SamplerInterface); isSampler {
-			sampleRows := p.selectMetricRows(layer.nextBucketAggregation.internalName+"count", rows)
-			sampleJson := layer.nextBucketAggregation.queryType.TranslateSqlResponseToJson(sampleRows, 0)
-			jsonWithOmittedSampler, err := p.layerToJSON(layerIdx+1, layers, rows)
-			if err != nil {
-				return nil, err
+		// sampler and filter are special
+		if !layer.nextBucketAggregation.DoesHaveGroupBy() {
+			// TODO: if filters/range/dateRange do something special
+			var selectedRows []model.QueryResultRow
+			switch layer.nextBucketAggregation.queryType.(type) {
+			case bucket_aggregations.FilterAgg:
+				selectedRows = p.selectMetricRows(layer.nextBucketAggregation.internalName+"_col_", rows)
+			case bucket_aggregations.SamplerInterface:
+				selectedRows = p.selectMetricRows(layer.nextBucketAggregation.internalName+"count", rows)
+			default:
+				return nil, fmt.Errorf("unexpected query type: %T", layer.nextBucketAggregation.queryType)
 			}
-			result[layer.nextBucketAggregation.name] = util.MergeMaps(context.Background(), sampleJson, jsonWithOmittedSampler, model.KeyAddedByQuesma)
-			return result, nil
-		}
-
-		// TODO: if filter/filters/range/dateRange do something special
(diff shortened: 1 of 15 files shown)


quesma/queryparser/pancake_transformer.go
```diff
@@ -137,6 +140,20 @@ func (a *pancakeTransformer) createLayer(previousAggrNames []string, childAggreg
 				result[0].layer.nextBucketAggregation = bucket
 				result[0].nextBucketAggregation = childAgg
 			} else {
+				// if both leaf optimizations are filter and second one doesn't have children we can treat second as metric
```
jakozaur, line 143: I don't like, but without that we generate two SQLs for sample dashboards.

quesma/queryparser/pancake_sql_query_generation.go
```diff
@@ -191,7 +192,39 @@ func (p *pancakeSqlQueryGenerator) generateBucketSqlParts(bucketAggregation *pan
...
+		for columnId, column := range metric.selectedColumns {
+			aliasedName := fmt.Sprintf("%s_col_%d", metric.internalName, columnId)
+			// Add if
+			var columnWithIf model.Expr
+			switch function := column.(type) {
+			case model.FunctionExpr:
+				if function.Name == "count" {
+					columnWithIf = model.NewFunction("countIf", whereClause)
+				} else if len(function.Args) == 1 {
+					columnWithIf = model.NewFunction(function.Name+"If", function.Args[0], whereClause)
```
pdelewski, line 213: Is this correct to do that (add `if`) for any function with one argument?

quesma/queryparser/pancake_sql_query_generation.go
```diff
@@ -214,6 +247,26 @@ func (p *pancakeSqlQueryGenerator) generateSelectCommand(aggregation *pancakeMod
...
+				}
+
+				if layerId+1 < len(aggregation.layers) {
+					addSelectColumns, err := p.generateLeafFilter(aggregation.layers[layerId+1], filter.WhereClause)
+					if err != nil {
+						return nil, false, err
+					}
+					selectColumns = append(selectColumns, addSelectColumns...)
+				}
+				break
```
pdelewski, line 265: Is this break needed?

quesma/queryparser/pancake_sql_query_generation.go
jakozaur: Seems so:
https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators#-if

quesma/queryparser/pancake_sql_query_generation.go
jakozaur: Yes, but this keeps me puzzled how to refactor it better.

Currently the logic is that filters are leaf, but they might be one more level to process. I escape the standard logic.

quesma/queryparser/pancake_sql_query_generation.go
pdelewski: I would add this link as a comment about rationale

nablaone: We can have a 'builder' function here.

quesma/queryparser/pancake_transformer.go
nablaone: This can be a part of the "pancake" optimization pipeline. Or it can be moved to the separate "optimizeSth()" function.

nablaone (approved):
Minor suggestions only. 

LGTM.

quesma/queryparser/pancake_json_rendering.go
```diff
@@ -24,6 +24,7 @@ func (p *pancakeJSONRenderer) selectMetricRows(metricName string, rows []model.Q
 		}
 		return []model.QueryResultRow{newRow}
 	}
+	logger.Error().Msgf("no rows in selectMetricRows %s", metricName)
```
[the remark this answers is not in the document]
jakozaur: Yes, it is later causing a panic.

PR #666, opened 2024-08-23
Combinators - parse/modify/render

nablaone:


jakozaur (commented):
I don't follow why we need it.

nablaone:
Evolutionary dead end.

PR #671, opened 2024-08-26
Ast cleanup

jakozaur:
Remove two AST types:
- deprecated `StringExpr`
- `MultiFunctionExpr`, it is a weird construct for quantiles, in fact this should be part of a function name. Got discussion with @nablaone


mieciu (approved):
Like it! 👍

pdelewski (approved):

PR #672, opened 2024-08-26
FROM clause resolution is part of pipeline

nablaone:
This is the first part of removing the `table` dependency from the query parser.  In this part physical table name is resolved in the pipeline. 

The next PR will cover timestamp/date fields ,


jakozaur (approved):
Love the direction, but would love if someone else gave it a try.

pdelewski:
Great Job! I believe it's the right direction

PR #674, opened 2024-08-27
Move aliases and types to schema overrides

mieciu:
Highlights:
* Moving `TypeMappings` and `Aliases` form `IndexConfiguration` to `SchemaConfiguration` (what used to be called `static-schema`, I'm also changing that to `schemaOverrides`
* Cleanup all the deprecated code logic
* There's more to come with the deprecated field changes, so you might see few line commented out here and there as I didn't want this PR to grow too much and conflict with some @nablaone's recent work

**This might look like only configuration syntax change on the surface**, but it actually aims to reduce the usage of configuration and make the code refer to schema instead.
image

### DIFF
--- a/examples/kibana-sample-data/quesma/config/local-dev.yaml
+++ b/examples/kibana-sample-data/quesma/config/local-dev.yaml
@@ -31,34 +31,44 @@ processors:
       indexes:
         kibana_sample_data_ecommerce:
           enabled: true
-          mappings:
-            products.manufacturer: "text"
-            products.product_name: "text"
-            geoip.location: "geo_point"
-            category: "text"
-            manufacturer: "text"
+          schemaOverrides:
+            fields:
+              "geoip.location":
+                type: geo_point
+              "products.manufacturer":
+                type: text
+              "products.product_name":
+                type: text
+              category:
+                type: text
+              manufacturer:
+                type: text
         kibana_sample_data_flights:
           enabled: true
-          mappings:
-            DestLocation: "geo_point"
-            OriginLocation: "geo_point"
+          schemaOverrides:
+            fields:
+              "DestLocation":
+                type: geo_point
+              "OriginLocation":
+                type: geo_point
         logs-generic-default:
           enabled: true
-          static-schema:
+          schemaOverrides:
             fields:
+              "@timestamp":
+                type: date
+                isTimestampField: true
+              timestamp:
+                type: alias
+                targetColumnName: "@timestamp"
               message:
-                type: "text"
-              host.name:
-                type: "keyword"
+                type: text
               service.name:
                 type: "keyword"
               source:
                 type: "keyword"
               severity:
                 type: "keyword"
-              severity_alias:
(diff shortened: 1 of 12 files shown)


quesma/clickhouse/parser.go
```diff
@@ -146,6 +146,8 @@ func SchemaToColumns(schemaMapping *schema.Schema, nameFormatter TableColumNameF
 			fType = "Nullable(DateTime64)"
 		case schema.TypeDate.Name:
 			fType = "Nullable(Date)"
+			// TODO if someone sets `type: date` to a field in schemaOverrides AND this is a timestamp field for which we have dedicated logic (use DateTime64 + add DEFAULT now64())
```
jakozaur, line 149: NOT, but TODO are mostly used in form `TODO:`

jakozaur: Not sure also if I understand `TODO`, they should be in form what is to be fixed and explanation.

quesma/clickhouse/table.go
```diff
@@ -21,8 +21,8 @@ type Table struct {
 	Config       *ChTableConfig
 	Created      bool // do we need to create it during first insert
 	indexes      []IndexStatement
-	aliases      map[string]string //deprecated
-	// we should use aliases directly from configuration, not store them here
+	aliases      map[string]string //deprecated TODO looks like we've reached the point where this becomes a problem
```
jakozaur, line 24: E.g. `TODO: Remove before first stable release. Use field with type alias instead.

jakozaur (approved):
Minor comments, someone else should also reviewed it.

examples/kibana-sample-data/quesma/config/local-dev.yaml
```diff
@@ -31,34 +31,41 @@ processors:
...
-          static-schema:
+          schemaOverrides:
             fields:
+              timestamp:
+                type: alias
+                targetColumnName: "@timestamp"
               message:
-                type: "text"
-              host.name:
-                type: "keyword"
```
avelanarius, line 52: Was there some problem with `host.name` that you decided to delete this?

avelanarius (approved):

examples/kibana-sample-data/quesma/config/local-dev.yaml
mieciu: it just slipped unnoticed :) bringing it back

examples/kibana-sample-data/quesma/config/local-dev.yaml
```diff
@@ -31,34 +31,43 @@ processors:
...
+              timestamp:
+                type: alias
+                targetColumnName: "@timestamp"
               message:
-                type: "text"
-              host.name:
+                type: text
+              "host.name":
                 type: "keyword"
               service.name:
```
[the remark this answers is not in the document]
mieciu: good catch!

pdelewski (approved):

PR #684, opened 2024-08-28
treewide: Pass QuesmaConfiguration by pointer, not by value

avelanarius:
While working on #683, I noticed that the code in `process()` in `Statistics` modified `QuesmaConfiguration` struct. However, since `QuesmaConfiguration` was passed by value, not by pointer, this modification essentially did nothing.

That (attempted) modification of global state is not that clean (fixed in the second commit), but **this highlighted an issue**: the entire codebase always passes `QuesmaConfiguration` by value (and this is quite a big struct!), not by pointer. Apart from potential performance penalty (but did not profile it), this also makes it not possible to change the configuration in the runtime (since everyone is using their own copy).

Fix the issue by modifying all uses of `QuesmaConfiguration` to pass it as a pointer.

### DIFF
--- a/quesma/ab_testing/sender/coordinator.go
+++ b/quesma/ab_testing/sender/coordinator.go
@@ -22,7 +22,7 @@ type SenderCoordinator struct {
 	enabled bool
 }
 
-func NewSenderCoordinator(cfg config.QuesmaConfiguration) *SenderCoordinator {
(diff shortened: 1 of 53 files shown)


avelanarius:
cc @mieciu - I hope this does not cause many conflicts with your ongoing work around configuration (if it does, this PR can wait)

jakozaur (approved):

mieciu (approved):
nice👌

PR #685, opened 2024-08-29
Store field type in map attribute column

pdelewski:
At some point we introduced `attributes_string_type` to not lose already inferred field type. 
With refactoring to maps we lost that. Now it's time to bring it back so `attribute` map has been extended
from `Map(String,String)` to `Map(String,Array(String))`

image


avelanarius:
Have you considered changing the type to `Map(String, Tuple(String, String))`? It would be more strongly typed and maybe has some performance/usability benefits?

pdelewski:
> Have you considered changing the type to `Map(String, Tuple(String, String))`? It would be more strongly typed and maybe has some performance/usability benefits?

Yes, that's another option. For now we know that `Map(String,Array(String)` works but I can check Map(String, Tuple(String,String) to check how much work is required to do so.

pdelewski:
Few solutions considered:
- Map(String,Array(String))
- Map(String,Tuple(String,String))
- separate maps for values and types Map(String,String)

pdelewski:
Closing due to https://github.com/QuesmaOrg/quesma/pull/691

PR #687, opened 2024-08-29
Fix `stats`/`quantiles` metric aggr

trzysiek:


quesma/testdata/aggregation_requests.go
```diff
@@ -4473,7 +4548,123 @@ var AggregationTests = []AggregationTestCase{
 				`WHERE toUnixTimestamp64Milli("timestamp")>=1.709932426749e+12 ` +
 				`AND toUnixTimestamp64Milli("timestamp")<=1.711228426749e+12`,
 		},
-		ExpectedPancakeSQL: "TODO",
+		ExpectedPancakeSQL: `
+			SELECT count(*) AS "aggr__sample__count",
+			  quantiles(0.050000)("bytes_gauge") AS "quantile_5" AS
```
jakozaur, line 4553: Do we really need this two `AS`? Maybe in next, but can we simplify one of the AS

jakozaur (approved):
Great work, one improvement, can be done in different PR.

Issue #307, opened 2024-06-11
`Date_histogram` doesn't work too well with timezones


trzysiek:
We'll need to think how to tackle time zones, in both Elastic, and Clickhouse.
How to reproduce:
I changed `docker/log-generator/logger`'s code for timestamp to:
```
stamps := []time.Time{
  time.Date(2021, 1, 2, 0, 0, 0, 0, time.UTC),
  time.Date(2021, 1, 1, 23, 50, 0, 0, time.UTC),
  time.Date(2021, 1, 1, 22, 55, 0, 0, time.UTC),
  time.Date(2021, 1, 2, 0, 55, 0, 0, time.UTC),
  time.Date(2021, 1, 2, 1, 55, 0, 0, time.UTC),
}
...
"@timestamp":   stamps[rand.Intn(len(stamps))].Format("2006-01-02T15:04:05.999Z"),
```
Then, there's discrepancy between our and Elastic's responses for query at the bottom. Ratio between hits in those 2 buckets is very different.
Elastic returns
![Screenshot 2024-06-11 at 11 00 40](https://github.com/QuesmaOrg/quesma/assets/5407146/dee28719-22a3-416a-b374-a1064c1c1766)
And we do:
![Screenshot 2024-06-11 at 11 03 59](https://github.com/QuesmaOrg/quesma/assets/5407146/c4904294-f941-454e-bb1a-de8d4fd9b496)

Query:
```
POST /logs-generic-default/_search
{
    "_source": {
        "excludes": []
    },
    "aggs": {
        "0": {
            "date_histogram": {
                "fixed_interval": "1d",
                "field": "@timestamp",
                "min_doc_count": 1,
                "time_zone": "Europe/Warsaw"
            }
        }
    },
    "fields": [
        {
            "field": "@timestamp",
            "format": "date_time"
        }
    ],
    "query": {
        "bool": {
            "filter": [
                {
                    "range": {
                        "@timestamp": {
                            "format": "strict_date_optional_time",
                            "gte": "2020-03-11T04:08:33.783Z",
                            "lte": "2024-06-11T03:08:33.783Z"
                        }
                    }
                }
            ],
            "must": [],
            "must_not": [],
            "should": []
        }
    },
    "runtime_mappings": {},
    "script_fields": {},
    "size": 0,
    "stored_fields": [
        "*"
    ],
    "track_total_hits": true
}
```

jakozaur:
Fixed.

PR #688, opened 2024-08-29
Fix timezone differences in `date_histogram`

trzysiek:
https://github.com/QuesmaOrg/quesma/issues/307

edit: one case is still wrong... Comment below `(that's also why we don't need to worry about the field being in different timezone in Clickhouse, it's translated to `UTC` anyway)` works in the second case, need to handle it ourselves in the first one.

For `date_histogram` we have 2 ways we generate SQLs:

a) for longer intervals: e.g. 1 month, we group by using this function `toStartOfMonth(timestamp)`. When taking timezones into consideration, it becomes `toStartOfMonth(toTimeZone(timestamp, "timezone_name"))`.
Example: We have such a simple table
Screenshot 2024-09-09 at 19 00 43

Before, such a request below returned what it returns now if we omit `time_zone` parameter, so `"key_as_string": "2023-12-01T00:00:00.000"`
But in Warsaw time, it's `2024-01-01 00:05`, so now we return good, new-year bucket.

Screenshot 2024-09-09 at 18 59 27

b) for shorter intervals, like the one in the issue https://github.com/QuesmaOrg/quesma/issues/307 ,
`toUnixTimestamp64Milli("@timestamp") / 3600000` becomes `toUnixTimestamp64Milli("@timestamp")+timeZoneOffset(toTimezone("@timestamp",'Europe/Warsaw'))*1000) / 3600000`
Here the SQL is a bit more complex, but I think it needs to be, as this `toUnixTimestamp` function always operates on `UTC`, translates different timezones to `UTC`, etc, so we need to add the offset ourselves.
(that's also why we don't need to worry about the field being in different timezone in Clickhouse, it's translated to `UTC` anyway)

You can see below that the answers now match those from Elastic, from https://github.com/QuesmaOrg/quesma/issues/307 , the ratio of `doc_count` for days is 1-4, not 2-3 as before this fix.
Screenshot 2024-09-09 at 19 12 20
in pancakes it works too
Screenshot 2024-09-09 at 19 16 43

### DIFF
--- a/quesma/testdata/clients/kunkka.go
+++ b/quesma/testdata/clients/kunkka.go
@@ -21,8 +21,7 @@ var KunkkaTests = []testdata.AggregationTestCase{
 				"0": {
 					"date_histogram": {
 						"field": "@timestamp",
-						"calendar_interval": "1h",
-						"time_zone": "Europe/Warsaw"
+						"calendar_interval": "1h"
 					},
 					"aggs": {
 						"1": {
@@ -347,8 +346,8 @@ var KunkkaTests = []testdata.AggregationTestCase{
 									"doc_count": 0
 								},
 								"doc_count": 2,
-								"key": 1718794800000,
-								"key_as_string": "2024-06-19T11:00:00.000"
+								"key": 1718787600000,
+								"key_as_string": "2024-06-19T09:00:00.000"
 							},
 							{
 								"1": {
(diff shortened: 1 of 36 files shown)


quesma/testdata/clients/kunkka.go
```diff
@@ -347,8 +346,8 @@ var KunkkaTests = []testdata.AggregationTestCase{
 									"doc_count": 0
 								},
 								"doc_count": 2,
-								"key": 1718794800000,
-								"key_as_string": "2024-06-19T11:00:00.000"
+								"key": 1718787600000,
+								"key_as_string": "2024-06-19T09:00:00.000"
```
trzysiek, line 350: Here and in a ton of other tests: if it's 11am bucket like here, in Warsaw time, then we returned 11:00, but it displayed bar at 13:00, because Kibana always interprets times in UTC (11 UTC = 13 Warsaw). Now 9:00 UTC = 11:00 Warsaw and it's good.

jakozaur (approved):
Somehow looks good, but prefer to have this logic in quesma instead.

PR #694, opened 2024-09-01
Return proper response for `significant_terms`, not default from `terms`

trzysiek:
`significant_terms` return `doc_count`, not `sum_other_doc_count` like terms.
Fixed here.


jakozaur (approved):
Amazing!

PR #698, opened 2024-09-02
Ingest - fixes

nablaone:
- **Return reasonable type in case of null**
- **Handle Nullable. Concurrency hack. Logging.**

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -468,12 +468,34 @@ func (lm *LogManager) generateNewColumns(
 	attrKeys := getAttributesByArrayName(AttributesKeyColumn, attrsMap)
 	attrTypes := getAttributesByArrayName(AttributesValueType, attrsMap)
 	var deleteIndexes []int
+
+	// HACK Alert:
+	// We must avoid altering the table.Cols map and reading at the same time.
+	// This should be protected by a lock or a copy of the table should be used.
+	//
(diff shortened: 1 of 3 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -474,12 +474,34 @@ func (lm *LogManager) generateNewColumns(
...
+		columnType := ""
+		// Array and Map are not Nullable
+		if strings.Contains(attrTypes[i], "Array") || strings.Contains(attrTypes[i], "Map") {
+			columnType = attrTypes[i]
+		} else {
+			columnType = fmt.Sprintf("Nullable(%s)", attrTypes[i])
+		}
+		alterTable := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN IF NOT EXISTS \"%s\" %s", table.Name, attrKeys[i], columnType)
+
+		newColumns[attrKeys[i]] = &Column{Name: attrKeys[i], Type: NewBaseType(attrTypes[i]), Modifiers: "Nullable"}
```
pdelewski, line 498: For now, we always adding `Nullable` modifiers which is inconsistent with above logic

pdelewski:
Overall, LGTM :)

pdelewski (approved):

quesma/clickhouse/clickhouse.go
nablaone: Right. I  will fix it before a merge.

PR #699, opened 2024-09-02
Alter column heuristic simplification

pdelewski:
Promotes field to column when it is present more or equal `fieldFrequency` times in last `alwaysAddColumnLimit`

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -34,7 +34,8 @@ const (
 	// Above this number of columns we will use heuristic
 	// to decide if we should add new columns
 	alwaysAddColumnLimit  = 100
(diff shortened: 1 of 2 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -34,7 +34,8 @@ const (
 	// Above this number of columns we will use heuristic
 	// to decide if we should add new columns
 	alwaysAddColumnLimit  = 100
-	AlterColumnUpperLimit = 1000
+	alterColumnUpperLimit = 1000
+	fieldFrequency        = 10
```
pdelewski, line 38: Arbitrarily chosen value, we can change that according to results of our empirical tests

avelanarius (approved):

PR #700, opened 2024-09-02
FullTextFields config parameter removal

nablaone:
This PR: 
- `FullTextFields` config parameter  is removed 
- schema type is used to determine which field is a 'full text' searchable


pdelewski:
@nablaone To clarify, Am I correct that now, all fields that are `Strings` and not `keywords` will be treaded as `fulltext` fields?

quesma/schema/registry.go
```diff
@@ -150,7 +150,7 @@ func (s *schemaRegistry) populateSchemaFromTableDefinition(definitions map[strin
 					fields[propertyName] = Field{PropertyName: propertyName, InternalPropertyName: FieldName(column.Name), Type: quesmaType}
 				} else {
 					logger.Debug().Msgf("type %s not supported, falling back to text", column.Type)
```
avelanarius, line 152:
```suggestion
					logger.Debug().Msgf("type %s not supported, falling back to keyword", column.Type)
```

avelanarius: (per change 1 line below)

mieciu (approved):
One thing @avelanarius caught, other than that LGTM :)

PR #701, opened 2024-09-02
Move IgnoredFields into SchemaOverrides, make it work properly with ingest

avelanarius:
This is a part of our recent effort of finalizing a new stable configuration. Previously `IgnoredFields` was a part of `IndexConfiguration`, looking like this:

```yaml
test_index:
  ignoredFields:
    myField: true
    anotherField: true
```

However, with the recent improvements to `SchemaConfiguration`, now it makes more sense to configure it under `schemaOverrides`, like this:

```yaml
test_index:
  schemaOverrides:
    fields:
      myField:
        ignored: true
      anotherField:
        ignored: true
```

This PR moves this configuration from `ignoredFields` (`IndexConfiguration`) to `schemaOverrides` (`SchemaConfiguration`).

Moreover, `ignoredFields` did not work correctly with Quesma's ingest, as ignored fields were only ignored in the query path. This PR adds necessary checks also to the ingest path.

Unfortunately, the code is not very clean with regards to handling of `.` vs `::` - I have tried cleaning up the situation, but that required some larger refactors which would make this PR too big...

### DIFF
--- a/quesma/jsonprocessor/ingest.go
+++ b/quesma/jsonprocessor/ingest.go
@@ -5,26 +5,44 @@ package jsonprocessor
 import (
 	"quesma/quesma/config"
 	"quesma/quesma/types"
+	"strings"
 )
 
 type IngestTransformer interface {
 	Transform(document types.JSON) (types.JSON, error)
 }
 
-type ingestTransformer struct {
+type flattenMapTransformer struct {
 	separator string
 }
 
-func (t *ingestTransformer) Transform(document types.JSON) (types.JSON, error) {
+func (t *flattenMapTransformer) Transform(document types.JSON) (types.JSON, error) {
 	return FlattenMap(document, t.separator), nil
 }
 
-// right now all transformers are the same, but we can add more in the future
-func IngestTransformerFor(table string, cfg *config.QuesmaConfiguration) IngestTransformer {
+type removeFieldsTransformer struct {
+	fields []config.FieldName
+}
 
+func (t *removeFieldsTransformer) Transform(document types.JSON) (types.JSON, error) {
+	for _, field := range t.fields {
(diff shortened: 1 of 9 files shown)


quesma/jsonprocessor/ingest.go
```diff
@@ -5,26 +5,44 @@ package jsonprocessor
...
-// right now all transformers are the same, but we can add more in the future
-func IngestTransformerFor(table string, cfg *config.QuesmaConfiguration) IngestTransformer {
+type removeFieldsTransformer struct {
+	fields []config.FieldName
+}
 
+func (t *removeFieldsTransformer) Transform(document types.JSON) (types.JSON, error) {
+	for _, field := range t.fields {
+		delete(document, field.AsString())
+		delete(document, strings.Replace(field.AsString(), ".", "::", -1))
```
avelanarius, line 30: This was required to correctly handle both:
```
{"ecs.version": "8.0.0"}
```
and
```
{"ecs": {"version": "8.0.0"}}
```

mieciu (commented):
Before I proceed with the further review - I'm a little concerned about bringing the special-use `ignored` field type.
IMHO types in this case should remain types, and `ignored` would be a better candidate for a separate field property, e.g.:
image

WDYT? Also curious what others think about this @QuesmaOrg/engineering

pdelewski:
> Before I proceed with the further review - I'm a little concerned about bringing the special-use `ignored` field type. IMHO types in this case should remain types, and `ignored` would be a better candidate for a separate field property, e.g.: image
> 
> WDYT? Also curious what others think about this @QuesmaOrg/engineering

Make sense for me

mieciu (approved):
LGTM :shipit:

PR #704, opened 2024-09-02
`timestampField` configuration field removal

nablaone:
Changes: 
- remove `timestampField` configuration parameter
-  `@timestamp` field is resolved in a query transformation pipeline now


quesma/model/typical_queries/hits.go
```diff
@@ -29,6 +29,7 @@ type Hits struct {
 	addScore       bool // true <=> we add hit.Score field to the response (whose value is always 1)
 	addVersion     bool // true <=> we add hit.Version field to the response (whose value is always 1)
 	indexName      string
+	useDefaultID   bool
```
mieciu, line 32: I think it now makes sense to place a little comment explaining what the default ID is in this context

quesma/clickhouse/table_discovery.go
```diff
@@ -189,14 +190,11 @@ func (td *tableDiscovery) autoConfigureTables(tables map[string]map[string]strin
...
-		} else {
-			configuredTables[table] = discoveredTable{table, columns, config.IndexConfiguration{}, comment, createTableQuery}
-		}
+		configuredTables[table] = discoveredTable{table, columns, config.IndexConfiguration{}, comment, createTableQuery, maybeTimestampField}
+
 	}
-	for tableName, conf := range configuredTables {
-		autoDiscoResults.WriteString(fmt.Sprintf("{table: %s, timestampField: %s}, ", tableName, conf.config.GetTimestampField()))
+	for tableName := range configuredTables {
+		autoDiscoResults.WriteString(fmt.Sprintf("{table: %s}, ", tableName))
```
mieciu, line 197: As this is auto discovery mode I think it's still worth printing TS field.

quesma/model/typical_queries/hits.go
```diff
@@ -132,12 +133,19 @@ func (query Hits) addAndHighlightHit(hit *model.SearchHit, resultRow *model.Quer
...
+
 func (query Hits) computeIdForDocument(doc model.SearchHit, defaultID string) string {
-	tsFieldName, err := query.table.GetTimestampFieldName()
-	if err != nil {
+
+	if query.useDefaultID {
 		return defaultID
 	}
 
+	tsFieldName := model.TimestampFieldName
```
mieciu, line 147: As we've chatted on Zoom, this isn't correct as it always takes `@timestamp`

quesma/model/typical_queries/hits.go
nablaone: It relies on the presence of a timestamp field now.

quesma/model/typical_queries/hits.go
nablaone: It is fixed now.

mieciu (approved):
LGTM now!

PR #706, opened 2024-09-02
[after pancake-only] Remove `Facets` entity from code

trzysiek:
We had a hot performance path for requests which looked like `facets`. After pancakes, it's probably no longer necessary, thus we try to get rid of that code.

Seems done, one test left to update. + I'll have to test it manually.


jakozaur:
It would be great if we commit some changes, but the facets were already removed by:
https://github.com/QuesmaOrg/quesma/pull/760

jakozaur:
No longer relevant.

PR #710, opened 2024-09-03
Base visitor - process `LIMIT BY` clause - fix

nablaone:


jakozaur (approved):
I believe we stopped using limit by in new pancake model. Happy to merge, but likely will remove it soon.

PR #711, opened 2024-09-03
Add a user-friendly error message when a column is missing.

nablaone:
Instead of a generic database error message, we should display something more helpful. 

Screenshot 2024-09-03 at 17 05 22


nablaone:
This is a naive solution. It will not work for queries like `select 1 as foo from bar where foo > 1`.

PR #718, opened 2024-09-05
One table to store them all 

nablaone:
This PR adds ability to store indices into a single table.  This is configured per index.  Here is an example:

```
        logs-elastic_agent-default:
          useSingleTable: true
        logs-elastic_agent.filebeat-default:
          useSingleTable: true
        logs-elastic_agent.metricbeat-default:
          useSingleTable: true
```

Changes:
-  Ingest path. Store data and manage the `quesma_all_logs` table. Manage virtual table definition for enabled indices.
- Query path. Rewrite queries for enabled indices.  
- UI. Add visualization: fields of enabled indices vs `quesma_all_logs` table columns 
- Persistent schema. Table definitions for these indices are stored in elastic search. 

Notice:
1. Data is stored into `quesma_all_table`. This is not configurable at this moment.
2. Putting a mapping does not alter the database. 
3. Table definition can be virtual now. This is a kind of view on `quesma_all_logs` table. Better name is needed here.
4. There is a `. -> ::` compensation added (see `handleDottedTColumnNames`) It's disabled by default.  @pdelewski we should discus this aspect.
 

Docker dashboard is "working" (with `handleDottedTColumnNames` compensation enabled): 

Screenshot 2024-09-16 at 15 37 56


quesma/clickhouse/table_discovery.go
```diff
@@ -48,16 +51,18 @@ type tableDiscovery struct {
...
 	result := &tableDiscovery{
-		cfg:              cfg,
-		dbConnPool:       dbConnPool,
-		tableDefinitions: &tableDefinitions,
-		forceReloadCh:    make(chan chan<- struct{}),
+		cfg:                 cfg,
+		dbConnPool:          dbConnPool,
+		tableDefinitions:    &tableDefinitions,
+		forceReloadCh:       make(chan chan<- struct{}),
+		virtualTableStorage: virtualTablesDB,
```
pdelewski, line 65: Maybe finally, we could merge `tableDefinitions` and `virtualTablesDB` or hide them as common interface

quesma/ingest/processor.go
```diff
@@ -621,17 +651,63 @@ func (ip *IngestProcessor) processInsertQuery(ctx context.Context,
...
+
+func (ip *IngestProcessor) processInsertQueryInternal(ctx context.Context, tableName string,
+	jsonData []types.JSON, transformer jsonprocessor.IngestTransformer,
+	tableFormatter TableColumNameFormatter, isVirtualTable bool) error {
+	statements, err := ip.processInsertQuery(ctx, tableName, jsonData, transformer, tableFormatter, isVirtualTable)
 	if err != nil {
 		return err
 	}
+
+	var logVirtualTableDDL bool // maybe this should be a part of the config or sth
```
pdelewski, line 692: hmm, seems that this variable is not initialized

pdelewski:
I merged these changes with https://github.com/QuesmaOrg/quesma/pull/767 and it seems that it works. I don't see interferences between these two features

quesma/quesma/schema_transformer.go
```diff
@@ -541,6 +627,7 @@ func (s *SchemaCheckPass) Transform(queries []*model.Query) ([]*model.Query, err
 		{TransformationName: "GeoTransformation", Transformation: s.applyGeoTransformations},
 		{TransformationName: "ArrayTransformation", Transformation: s.applyArrayTransformations},
 		{TransformationName: "MapTransformation", Transformation: s.applyMapTransformations},
+		{TransformationName: "DottedColumnNames", Transformation: s.handleDottedTColumnNames},
```
pdelewski, line 630: That's ok for now, but finally we should get rid of that

pdelewski:
I'm ok with merging it as it is to not block your work

pdelewski (approved):

quesma/single_table/const.go
```diff
@@ -0,0 +1,52 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+package single_table
+
+import (
+	"database/sql"
+	"quesma/logger"
+)
+
+const TableName = "quesma_all_logs"
```
mieciu, line 10:
```suggestion
const TableName = "quesma_common_table"
```
how about this?

nablaone: Looks good. I'll refactor it. This feature is called `Common Table` now.

PR #722, opened 2024-09-06
Implicitly enable optimizers when config is there

mieciu:
Similarly to indices (ref: https://github.com/QuesmaOrg/quesma/pull/682), optimizers are implicitly enabled when their configuration is in place. There's an option to leave config and make them have no effect by using `disabled` boolean flag.

Old reality:
```
optimizers:
  pancake:
    Enabled: true
    properties:
      mode: "apply"
```

Becomes:
```
optimizers:
  pancake:
    properties:
      mode: "apply"
```
 or optionally:
 ```
optimizers:
  pancake:
    disabled: false
    properties:
      mode: "apply"
```
FYSA @nablaone.


pdelewski (approved):

mieciu:
FYSA @jakozaur

PR #723, opened 2024-09-06
Nested field encoding, moving encoding into one clearly defined place #1

pdelewski:
The purpose of this PRs is to unify nested field encoding process, by doing this exactly in one place.
For this reason we have to move generation of `Create Table` string and `insert into` into one function.


mieciu (approved):
👍  way to go!

avelanarius (approved):

PR #631, opened 2024-08-09
A/B Testing - elastic 

nablaone:
This PR add:
- add support for comparison of  Elastic  vs Quesma responses
- add minor improvements related to storing the results 
- move alternative plan handling to a separate file 

Comparison with an Elastic or other external Quesma can be enabled by:
```
  some-index:
    enabled: true
    optimizers:
      elastic_ab_testing:
        enabled: true
        properties:
          mode: "alternative" 
```
Configuration will change in the future.

PR #724, opened 2024-09-06
Push Elastic's response from A/B testing to management console

avelanarius:
In the Quesma Managment Console we have "Elastic response" panel in the "Live tail" page and similar on the drilldown page. Those panels were filled in by the "EnableElasticsearchIngest: true" mode, which resends all requests to Elastic (even non-ingest!).

However in the upcoming PRs we will effectively remove "EnableElasticsearchIngest" mode. For the query path, it's now obsoleted by A/B testing with Elastic (see #631). But A/B testing with Elastic did not fill in the Quesma Management Console with responses from Elastic.

This PR improves A/B testing with Elastic to also send Elastic's response to the Quesma Management Console.

This is how the relevant panels look like now (previously they were empty):
Zrzut ekranu 2024-09-6 o 12 27 00
Zrzut ekranu 2024-09-6 o 12 26 21

### DIFF
--- a/quesma/quesma/search.go
+++ b/quesma/quesma/search.go
@@ -766,6 +766,14 @@ func (q *QueryRunner) postProcessResults(table *clickhouse.Table, results [][]mo
 	return geoIpTransformer.Transform(res)
 }
 
+func pushPrimaryInfo(qmc *ui.QuesmaManagementConsole, Id string, QueryResp []byte, startTime time.Time) {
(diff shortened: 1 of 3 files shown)


avelanarius:
(My primary motivation was that in the upcoming PRs I will effectively remove `EnableElasticsearchIngest` and I didn't want to break existing functionality, such as "Elasticsearch response" panels)

pdelewski (approved):

jakozaur (approved):

PR #725, opened 2024-09-06
nested fields encoding, refactoring ddl creation #2

pdelewski:
Propagating the return of type-safe columns as []CreateTableEntry and map[schema.FieldName]CreateTableEntry instead of strings.


mieciu (approved):
:shipit:

PR #726, opened 2024-09-06
Remove `EnableElasticsearchIngest`, operation modes

avelanarius:
## Remove `EnableElasticsearchIngest`

Remove `EnableElasticsearchIngest` option. This mode essentially always resent each request to ElasticSearch. Contrary to its name, it resent it for both query path and ingest path. This re-sending happened on a low-level - routing logic. It turns out this is too low-level for query and ingest needs:

### Query needs:
A/B testing with Elastic (a newer approach) is already implemented and achieves more than could be done at the low-level routing level (e.g. comparisons, only doing it for search endpoints, not other endpoints)

### Ingest needs:
`EnableElasticsearchIngest` primary use-case was double writes to Elastic and ClickHouse, but this flag was global, not configured per each index. In the upcoming PR we will reimplement double writes at a higher level, not at the routing level ("dumb" replaying of HTTP requests)

## Remove operation modes

Remove operation modes abstraction. Operation modes were modes
like `DualWriteQueryClickhouse`, `ProxyInspect`,
`DualWriteQueryClickhouseFallback`.

This commit removes them, because:
- Most of them did not work as advertised (for example no logic for verification in `DualWriteQueryClickhouseVerify`, dual writes were not actually turned on by default in `DualWrite*` modes, `DualWriteQueryClickhouseVerify` is now achieved with A/B testing)
- Modes are specified globally, not per index (all indexes share them)
- Mode describes query path and ingest path jointly, but in config_v2 they are separate things
- config_v2 actually only used `DualWriteQueryClickhouse` and `ProxyInspect`, all other modes were unused
- The upcoming new index configuration certainly will not use mode abstraction


quesma/quesma/config/config_v2.go
```diff
@@ -407,20 +407,20 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 		}
 	}
 	// Now determine Quesma final state with following heuristic
-	// if 2 pipelines with noop processor -> switch to proxy mode, ditch the whole config
-	// if one query, one ingest pipeline with noop processor -> switch to "dual-write-query-clickhouse" mode
+	// if 2 pipelines with noop processor -> switch to transparent proxy mode, ditch the whole config
+	// if one query, one ingest pipeline with noop processor -> switch to "Quesma" mode
```
mieciu, line 411: I think right now the whole logic is as simple as "if all configured processors are of type noop -> switch to transparent proxy mode"  :)

quesma/quesma/config/test_config.yaml
```diff
@@ -1,5 +1,5 @@
 # TEST CONFIGURATION
-mode: "dual-write-query-clickhouse"
+transparentProxy: false
```
mieciu, line 2: let's remove this from here and stick to the idea that any "mode" should be inferred from pipeline/processor configuration

avelanarius: This file is the "old" config (not config_v2) - that's the reason it's still here.

There is no `transparentProxy` is new config.

Maybe we should delete this old config file? (It's only used in tests of deserialization of old config)

mieciu: yea that's what I've thought 😉 I don't think we need that anymore, so go ahead with further removals

mieciu (changes requested):
left some comments here and there, once you address them I'm good with merging.

Nice work overall!

quesma/quesma/config/config_v2.go
avelanarius: Fixed in https://github.com/QuesmaOrg/quesma/pull/726/commits/87b5ea91fdd00d86b1c54d2ad327addb1c86e1e4

quesma/quesma/config/test_config.yaml
avelanarius: Migrated part of tests to test new config v2 instead of old config in this commit: https://github.com/QuesmaOrg/quesma/pull/726/commits/35b68ab80c79e2d10897d90fe61c46c6d3f0f2a9

There were some tests that tested adding Hydrolix connector via env variables, but it doesn't make that much sense with pipelines and connectors, therefore deleted those 2 tests. When I'll send PR with new index configuration, I'll add some more tests though.

mieciu (approved):
:shipit:

PR #728, opened 2024-09-06
A/B testing - start ASAP

nablaone:
A/B testing starts without any delay now.  The delay caused an 'it doesn't work' experience for fast devs.

### DIFF
--- a/quesma/ab_testing/collector/collector.go
+++ b/quesma/ab_testing/collector/collector.go
@@ -57,6 +57,10 @@ type InMemoryCollector struct {
 	// add  health state
 }
 
+func (r *InMemoryCollector) String() string {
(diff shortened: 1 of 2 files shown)


pdelewski (approved):

avelanarius:
Wouldn't a cleaner approach be to add `sendHealthMessage()` before the loop here: https://github.com/QuesmaOrg/quesma/blob/c1252dcc855a07bfa2056e28861148c42e6f335a/quesma/ab_testing/collector/collector.go#L145

avelanarius:
(collector immediately saying it's healthy vs collector user assuming that the collector is OK at start, if I understand it correctly)

nablaone:
Yes. We assume that it's born healthy. This part is a bit over-engineered right now.

PR #731, opened 2024-09-09
Refactoring - moving ingest field name encoding into one function

pdelewski:
This PR refactor ingest workflow and unifies ingest field encoding inside `processInsertQuery` function.
That's not finish, it's just a step forward

### DIFF
--- a/quesma/clickhouse/clickhouse.go
+++ b/quesma/clickhouse/clickhouse.go
@@ -524,7 +524,7 @@ type NonSchemaField struct {
 	Type  string
 }
 
-func generateNonSchemaStr(nonSchemaFields []NonSchemaField) string {
+func convertNonSchemaFieldsToString(nonSchemaFields []NonSchemaField) string {
 	if len(nonSchemaFields) <= 0 {
 		return ""
 	}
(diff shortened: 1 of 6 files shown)


quesma/clickhouse/clickhouse.go
```diff
@@ -518,42 +518,79 @@ func (lm *LogManager) generateNewColumns(
 	return alterCmd
 }
 
-func generateNonSchemaFieldsString(attrsMap map[string][]interface{}) (string, error) {
+// This struct contains the information about the columns that are part of the schema
```
mieciu, line 521:
```suggestion
// This struct contains the information about the columns that aren't part of the schema
```
😉

quesma/clickhouse/clickhouse.go
```diff
@@ -518,42 +518,79 @@ func (lm *LogManager) generateNewColumns(
...
+				value = nonSchemaFields[i].Type
+			} else {
+				value = nonSchemaFields[i].Value
+			}
+			if i > 0 {
+				nonSchemaStr += ","
+			}
+			nonSchemaStr += fmt.Sprintf("\"%s\":\"%s\"", nonSchemaFields[i].Key, value)
+		}
+		nonSchemaStr = nonSchemaStr + "}"
```
mieciu, line 552: Hmm would StringBuilder make this look a little better? I'm nor sure myself, I'll leave that up to you :)

mieciu (approved):
One note on StringBuilder, but if you don't feel like doing it within this PR, I'm fine with the code as is 🙂

quesma/clickhouse/clickhouse.go
pdelewski: Yes, good point

PR #732, opened 2024-09-09
Introducing internal Quesma type

mieciu:
We've found the `Type` name being very elusive in our codebase when working on schema persistence with @nablaone. 

This PR introduces `QuesmaType*` so that it becomes very clear in our codebase when we're referring to ClickHouse/Elastic types.

Additionally, `config.QuesmaConfiguration` is no longer used in SchemaRegistry  in favor of just index configuration.


nablaone (approved):
LGTM.

PR #733, opened 2024-09-09
Use Schema Registry in a query transformation pipeline 

nablaone:
We should use `schemaRegistry` everywhere to resolve indices and their fields.


mieciu (commented):
non-binding LGTM as I'd like someone else to look at this as well 👍

pdelewski (approved):

PR #735, opened 2024-09-09
Top metrics

jakozaur:
Implements `top_metrics`, which brings pancake to parity with previous processing and in some cases exceeding.

This also implements filters/range/dataRange for `top_hits`.

### DIFF
--- a/quesma/model/bucket_aggregations/dateRange.go
+++ b/quesma/model/bucket_aggregations/dateRange.go
@@ -196,6 +196,9 @@ func (query DateRange) DoesNotHaveGroupBy() bool {
 func (query DateRange) CombinatorGroups() (result []CombinatorGroup) {
 	for intervalIdx, interval := range query.Intervals {
 		prefix := fmt.Sprintf("range_%d__", intervalIdx)
+		if len(query.Intervals) == 1 {
+			prefix = ""
+		}
 		result = append(result, CombinatorGroup{
 			idx:         intervalIdx,
 			Prefix:      prefix,
@@ -228,4 +231,12 @@ func (query DateRange) CombinatorTranslateSqlResponseToJson(subGroup CombinatorG
 	}
 
 	return response
+}
+
+func (query DateRange) CombinatorSplit() []model.QueryType {
+	result := make([]model.QueryType, 0, len(query.Intervals))
+	for _, interval := range query.Intervals {
+		result = append(result, NewDateRange(query.ctx, query.FieldName, query.Format, []DateTimeInterval{interval}, query.SelectColumnsNr))
+	}
+	return result
 }
--- a/quesma/queryparser/pancake_json_rendering.go
+++ b/quesma/queryparser/pancake_json_rendering.go
@@ -40,7 +40,9 @@ func (p *pancakeJSONRenderer) selectMetricRows(metricName string, rows []model.Q
 	return
 }
 
-func (p *pancakeJSONRenderer) selectTopRows(topAggr *pancakeModelMetricAggregation, rows []model.QueryResultRow) (result []model.QueryResultRow) {
+// selectTopHitsRows: select columns for top_hits/top_metrics and rename them to original column names.
+// There is refactoring opportunity once we move completely to pancakes and remove re-name logic from this method.
+func (p *pancakeJSONRenderer) selectTopHitsRows(topAggr *pancakeModelMetricAggregation, rows []model.QueryResultRow) (result []model.QueryResultRow) {
 	for _, row := range rows {
 		var newCols []model.QueryResultCol
 		for _, col := range row.Cols {
@@ -231,7 +233,7 @@ func (p *pancakeJSONRenderer) layerToJSON(remainingLayers []*pancakeModelLayer, 
 		var metricRows []model.QueryResultRow
 		switch metric.queryType.(type) {
 		case metrics_aggregations.TopMetrics, metrics_aggregations.TopHits:
-			metricRows = p.selectTopRows(metric, rows)
+			metricRows = p.selectTopHitsRows(metric, rows)
 		default:
 			metricRows = p.selectMetricRows(metric.InternalNamePrefix(), rows)
 		}
--- a/quesma/queryparser/pancake_json_rendering.go
+++ b/quesma/queryparser/pancake_json_rendering.go
@@ -40,18 +40,18 @@ func (p *pancakeJSONRenderer) selectMetricRows(metricName string, rows []model.Q
 	return
 }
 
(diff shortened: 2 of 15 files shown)


quesma/queryparser/pancake_json_rendering.go
```diff
@@ -40,18 +40,18 @@ func (p *pancakeJSONRenderer) selectMetricRows(metricName string, rows []model.Q
 	return
 }
 
-func (p *pancakeJSONRenderer) selectTopHitsRows(topHits *pancakeModelMetricAggregation, rows []model.QueryResultRow) (result []model.QueryResultRow) {
+func (p *pancakeJSONRenderer) selectTopRows(topAggr *pancakeModelMetricAggregation, rows []model.QueryResultRow) (result []model.QueryResultRow) {
```
trzysiek, line 43: I think comment is needed what this function does. With `TopHits` in name it was more or less clear, now I don't think so.

quesma/queryparser/pancake_top_hits.go
```diff
@@ -14,9 +14,10 @@ func (p *pancakeSqlQueryGenerator) quotedLiteral(name string) model.LiteralExpr
 	return model.NewLiteral(strconv.Quote(name))
 }
 
-func (p *pancakeSqlQueryGenerator) generateSimpleTopHitsQuery(topHits *pancakeModelMetricAggregation,
+func (p *pancakeSqlQueryGenerator) generateSimpleTopQuery(topHits *pancakeModelMetricAggregation,
```
trzysiek, line 17: change `topHits` name

quesma/queryparser/pancake_top_hits.go
```diff
@@ -28,25 +29,41 @@ func (p *pancakeSqlQueryGenerator) generateSimpleTopHitsQuery(topHits *pancakeMo
...
-		Limit:       topHitsQueryType.Size,
+		OrderBy:     orderBy,
+		Limit:       size,
 	}
 
 	return resultQuery, nil
 }
 
-func (p *pancakeSqlQueryGenerator) generateTopHitsQuery(aggregation *pancakeModel,
+func (p *pancakeSqlQueryGenerator) generateTopQuery(aggregation *pancakeModel,
```
trzysiek, line 39: change `topHits` name, also maybe some comment here as well

quesma/queryparser/pancake_transformer.go
```diff
@@ -297,21 +300,23 @@ func (a *pancakeTransformer) findParentBucketLayer(layers []*pancakeModelLayer,
...
+func (a *pancakeTransformer) createTopHitAndTopMetricsPancakes(pancake *pancakeModel) (result []*pancakeModel, err error) {
 	for layerIdx, layer := range pancake.layers {
 		metricsWithoutTopHits := make([]*pancakeModelMetricAggregation, 0, len(layer.currentMetricAggregations))
-		for _, metric := range layer.currentMetricAggregations {
-			if _, isTopHits := metric.queryType.(metrics_aggregations.TopHits); isTopHits {
-				canOptimize := layerIdx == len(pancake.layers)-1 && len(layer.currentMetricAggregations) == 1
+		for metricIDx, metric := range layer.currentMetricAggregations {
+			switch metric.queryType.(type) {
+			case metrics_aggregations.TopMetrics, metrics_aggregations.TopHits:
+				canOptimize := layerIdx == len(pancake.layers)-1 && (len(layer.currentMetricAggregations) == 1 ||
```
nablaone, line 309: The condition is quite complex. Maybe we should add a comment with examples.

quesma/queryparser/pancake_transformer.go
```diff
@@ -297,21 +300,23 @@ func (a *pancakeTransformer) findParentBucketLayer(layers []*pancakeModelLayer,
...
 }
 
-func (a *pancakeTransformer) createTopHitPancakes(pancake *pancakeModel) (result []*pancakeModel, err error) {
+func (a *pancakeTransformer) createTopHitAndTopMetricsPancakes(pancake *pancakeModel) (result []*pancakeModel, err error) {
 	for layerIdx, layer := range pancake.layers {
 		metricsWithoutTopHits := make([]*pancakeModelMetricAggregation, 0, len(layer.currentMetricAggregations))
-		for _, metric := range layer.currentMetricAggregations {
-			if _, isTopHits := metric.queryType.(metrics_aggregations.TopHits); isTopHits {
-				canOptimize := layerIdx == len(pancake.layers)-1 && len(layer.currentMetricAggregations) == 1
+		for metricIDx, metric := range layer.currentMetricAggregations {
```
trzysiek, line 306:
```suggestion
		for metricIdx, metric := range layer.currentMetricAggregations {
```
to be consistent with naming everywhere

nablaone (approved):
Some comments will be helpful.
Overall LGTM.

quesma/model/bucket_aggregations/dateRange.go
```diff
@@ -196,6 +196,9 @@ func (query DateRange) DoesNotHaveGroupBy() bool {
 func (query DateRange) CombinatorGroups() (result []CombinatorGroup) {
 	for intervalIdx, interval := range query.Intervals {
 		prefix := fmt.Sprintf("range_%d__", intervalIdx)
+		if len(query.Intervals) == 1 {
```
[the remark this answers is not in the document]
jakozaur: So, it is a combinator approach that adds functions like `countIf` is incompatible with JOINs needed for `top_hits` and `top_metrics`.  I fixed it by flattening this typo of aggregations to one range and creating multiple SQLs.

However, than either needs to add special handling of column prefixes which is tricky in joins or simple skip prefixes if there is just one group. Since `filter` does not add prefix, I went with that approach.

quesma/queryparser/pancake_json_rendering.go
jakozaur: Will keep TopHits convention and add comment.

The re-name column logic could be simplified once we move completely to pancakes.

quesma/queryparser/pancake_top_hits.go
jakozaur: I will fix and add comment.

quesma/queryparser/pancake_top_hits.go
jakozaur: I will fix and add comment.

quesma/queryparser/pancake_transformer.go
jakozaur: Fixed.

quesma/queryparser/pancake_transformer.go
jakozaur: Broke it down to named variables.

PR #739, opened 2024-09-10
 Fix validation of single pipeline case

avelanarius:
In the `config_v2.go` we support a single pipeline use-case. The code permitted having a single pipeline for querying with processor: either `QuesmaV1ProcessorNoOp` or `QuesmaV1ProcessorQuery`. This commit fixes two things about it:

1. This validation is misleading - the user requested Quesma with only querying enabled, but actually ingest got also enabled. Therefore this commit will now reject query-only pipelines.

2. `if proc.Type == QuesmaV1ProcessorQuery` was incorrectly nested (it was inside `if proc.Type == QuesmaV1ProcessorQuery`), therefore the code was actually never executed.


mieciu (approved):
👍

PR #742, opened 2024-09-10
Geo in pancakes

jakozaur:
Fixes this panel:
Screenshot 2024-09-10 at 17 29 22

### DIFF
--- a/quesma/model/bucket_aggregations/geotile_grid.go
+++ b/quesma/model/bucket_aggregations/geotile_grid.go
@@ -30,9 +30,9 @@ func (query GeoTileGrid) TranslateSqlResponseToJson(rows []model.QueryResultRow,
 	}
 	var response []model.JsonMap
 	for _, row := range rows {
-		zoom := int64(row.Cols[0].Value.(float64))
-		x := int64(row.Cols[1].Value.(float64))
-		y := int64(row.Cols[2].Value.(float64))
+		zoom := int64(query.extractFloat(row.Cols[0].Value))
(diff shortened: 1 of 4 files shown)


quesma/model/bucket_aggregations/geotile_grid.go
```diff
@@ -44,6 +44,18 @@ func (query GeoTileGrid) TranslateSqlResponseToJson(rows []model.QueryResultRow,
...
+	case float64:
+		return valueTyped
+	case float32:
+		return float64(valueTyped)
+	default:
+		logger.ErrorWithCtx(query.ctx).Msgf("unexpected value type in geotile_grid aggregation response: %T", value)
+		return 0
+	}
+}
+
```
trzysiek, line 58: If a column is of `Nullable(Float)` type, it'll return `*float`. Will it always be just `Float` here?
Maybe use this function from `util/utils.go`?
```
func ExtractFloat64(value any) float64 {
	switch valueTyped := value.(type) {
	case float64:
		return valueTyped
	case *float64:
		return *valueTyped
	case float32:
		return float64(valueTyped)
	case *float32:
		return float64(*valueTyped)
	}
	logger.Error().Msgf("ExtractFloat64, value of incorrect type. Expected (*)float64, received: %v; type: %T", value, value)
	return -1
}
```

jakozaur: Fixed.

trzysiek (approved):

PR #743, opened 2024-09-10
Add "target" field to index configuration (part 1, bare minimum)

avelanarius:
When configuring indexes, users have a variety of use-cases, such as:
- querying from Elastic, but ingesting into ClickHouse
- querying from ClickHouse and ingesting into ClickHouse
- querying from Elastic, but ingesting into both Elastic and ClickHouse

The existing configuration was not flexible enough to encode those scenarios. This commit adds an additional required config option `target` that specifies the desired source/destination backend connector per each index. Since query and ingest processors are separate, this will allow in the future to have different preferences for query and ingest (not yet in this commit).

This added support in this commit is a bare-minimum support of the `target` config option. For example, it doesn't yet permit dual writes (specifying two targets) or having different configuration of ingest and query.

The actual query/ingest code is not yet changed in this PR and does not use the new `QueryTarget`, `IngestTarget` values. Currently the only permitted scenarios are: query+ingest into ES or query+ingest into CH, the first one being emulated by setting `Disabled=true`.

The following PRs will relax those constraints and enable more scenarios.

### DIFF
--- a/ci/quesma/config/ci-config.yaml
+++ b/ci/quesma/config/ci-config.yaml
@@ -23,27 +23,37 @@ processors:
     config:
       indexes:
         kibana_sample_data_ecommerce:
+          target: [ my-clickhouse-data-source ]
         kibana_sample_data_flights:
+          target: [ my-clickhouse-data-source ]
         kibana_sample_data_logs:
+          target: [ my-clickhouse-data-source ]
         logs-generic-default:
+          target: [ my-clickhouse-data-source ]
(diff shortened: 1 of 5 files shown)


avelanarius:
Example config files updated here: https://github.com/QuesmaOrg/quesma-examples/pull/22

mieciu (commented):
LGTM  👍

mieciu (approved):

PR #744, opened 2024-09-10
Fix last flight panel

jakozaur:
Fix the last flight panel previously we had fair chance of showing error there:
Screenshot 2024-09-10 at 18 39 08

Two fixes:
- if you do * query, even if there is no full-text field, we should default to true
- Implement `min_doc_count` for values different than zero, we also need to change default to `-1`

### DIFF
--- a/quesma/model/bucket_aggregations/date_histogram.go
+++ b/quesma/model/bucket_aggregations/date_histogram.go
@@ -9,6 +9,7 @@ import (
 	"quesma/kibana"
 	"quesma/logger"
 	"quesma/model"
+	"quesma/util"
 	"strconv"
 	"strings"
 	"time"
@@ -17,7 +18,7 @@ import (
 type DateHistogramIntervalType bool
 
 const (
-	DefaultMinDocCount                                      = 1
+	DefaultMinDocCount                                      = -1
 	DateHistogramFixedInterval    DateHistogramIntervalType = true
 	DateHistogramCalendarInterval DateHistogramIntervalType = false
 	defaultDateTimeType                                     = clickhouse.DateTime64
@@ -64,6 +65,9 @@ func (query *DateHistogram) TranslateSqlResponseToJson(rows []model.QueryResultR
 		)
 	}
 
+	// TODO: implement default  when query.minDocCount == DefaultMinDocCount
+	// all buckets between the first bucket that matches documents and the last one are returned
+
 	if query.minDocCount == 0 {
 		rows = query.NewRowsTransformer().Transform(query.ctx, rows)
 	}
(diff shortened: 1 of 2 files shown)


quesma/model/bucket_aggregations/date_histogram.go
```diff
@@ -91,6 +95,11 @@ func (query *DateHistogram) TranslateSqlResponseToJson(rows []model.QueryResultR
 		_, timezoneOffsetInSeconds := intervalStartNotUTC.Zone()
 		key -= int64(timezoneOffsetInSeconds * 1000) // seconds -> milliseconds
 
+		docCount := row.LastColValue()
+		if util.ExtractInt64(docCount) < int64(query.minDocCount) {
+			continue
+		}
```
trzysiek, line 101: maybe move to the top of `for`? Also use this `docCount` below in the response map

quesma/quesma/schema_transformer.go
```diff
@@ -386,6 +386,9 @@ func (s *SchemaCheckPass) applyFullTextField(indexSchema schema.Schema, query *m
 			if col.ColumnName == model.FullTextFieldNamePlaceHolder {
 
 				if len(fullTextFields) == 0 {
+					if (e.Op == "LIKE" || e.Op == "ILIKE") && model.AsString(e.Right) == "'%'" {
```
trzysiek, line 389: Maybe to avoid some subtle bugs?
```suggestion
					if (strings.ToUpper(e.Op) == "LIKE" || strings.ToUpper(e.Op) == "ILIKE") && model.AsString(e.Right) == "'%'" {
```

quesma/model/bucket_aggregations/date_histogram.go
```diff
@@ -17,7 +18,7 @@ import (
 type DateHistogramIntervalType bool
 
 const (
-	DefaultMinDocCount                                      = 1
+	DefaultMinDocCount                                      = -1
```
trzysiek, line 21: Why? It's confusing to me, as the default seems to be `1`

quesma/model/bucket_aggregations/date_histogram.go
```diff
@@ -64,6 +65,9 @@ func (query *DateHistogram) TranslateSqlResponseToJson(rows []model.QueryResultR
 		)
 	}
 
+	// TODO: implement default  when query.minDocCount == DefaultMinDocCount
+	// all buckets between the first bucket that matches documents and the last one are returned
```
trzysiek, line 69: I don't understand this TODO. AFAIK default is just `1` and what you've described is what happens when it's `0`, and it's already implemented.

quesma/model/bucket_aggregations/date_histogram.go
jakozaur: Now the default is to return all buckets from first hit till last one even if they are zeros. A lot of tests break b/c of that.

jakozaur:
Check docs or put 1. A lot of tests break.

trzysiek (approved):

PR #746, opened 2024-09-10
Dissecting ingest from `LogManager`

mieciu:
As in the description, moving ingest-related code to separate package, reducing the responsibilities of `LogManager` object.

Because currently ingest depends on `SchemaRegistry`, it was  impossible (cyclic dependencies) to add `clickhouse.Table` type to LogManager. Separating that will enable us to have all the table-related operations done via the registry and gradually migrate to that direction.

Additionally, this refactor gave birth to `IngestProcessor`.


pdelewski:
As discussed, this seems step in the right direction

pdelewski (approved):

PR #748, opened 2024-09-11
Adding Field encoding logic

pdelewski:
Field encoding logic is common and will be used both by ingest and query, so for now I decided to add this logic to `util` package

### DIFF
--- a/quesma/util/utils.go
+++ b/quesma/util/utils.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"unicode"
 )
 
 type JsonMap = map[string]interface{}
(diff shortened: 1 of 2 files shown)


quesma/util/utils.go
```diff
@@ -790,3 +791,28 @@ func Stringify(v interface{}) string {
 	isInsideArray := false
 	return stringifyHelper(v, isInsideArray)
 }
+
+const timestampFieldName = "@timestamp"
+
+func replaceNonAlphabetic(str string) string {
+	runes := []rune(str)
+	for i, r := range runes {
+		if !unicode.IsLetter(r) && !unicode.IsDigit(r) {
```
avelanarius, line 800: Shouldn't we use something more strict than `unicode.IsLetter`? Provided I understand the documentation of `IsLetter` correctly, this check will permit some non-ASCII letter like: "ʄ", "ѩ", "ά", "ﬗ" (characters from this Unicode group: https://www.compart.com/en/unicode/category/Ll).

Shouldn't the check be: "is ASCII letter"?

avelanarius: See: https://go.dev/play/p/zBA3VyU5dr0

The same case seems to be with `IsDigit` check, as "۴" or "൫" are valid Unicode digits.

pdelewski: Yes, good point

quesma/util/utils_test.go
```diff
@@ -821,3 +821,26 @@ func TestExtractInt64(t *testing.T) {
...
+	tests := []struct {
+		field string
+		want  string
+	}{
+		{"@timestamp", "@timestamp"},
+		{"timestamp", "timestamp"},
+		{"9field", "_9field"},
+		{"field9", "field9"},
+		{"field::", "field__"},
+		{"host.name", "host_name"},
```
avelanarius, line 835: Nit: you could add a couple instances of "strange Unicode characters" cases here.

avelanarius (approved):

quesma/util/utils_test.go
pdelewski: ok, Will do that in next PR(s)

PR #751, opened 2024-09-11
Removing bizarre import cycles pt 1

mieciu:
This PR aims to remove some odd dependencies which still prevents us from adding "physical" table information to schema registry.

* removal of `"quesma/elasticsearch"` import from the ClickHouse package
* Remove deprecated `ResolveTableName`
* `ResolveIndexes` renamed to more appropriate `ResolveIndexPattern`
* further cleanup, around the IngestProcessor 😉


quesma/clickhouse/clickhouse.go
```diff
@@ -132,54 +131,39 @@ func (lm *LogManager) Close() {
...
 		return nil, err
 	}
 
 	results = make([]string, 0)
-	if strings.Contains(patterns, ",") {
-		for _, pattern := range strings.Split(patterns, ",") {
-			if pattern == elasticsearch.AllIndexesAliasIndexName || pattern == "" {
+	if strings.Contains(pattern, ",") {
+		for _, pattern := range strings.Split(pattern, ",") {
+			if pattern == "_all" || pattern == "" {
```
pdelewski, line 146: nitpick: As this `_all` is used at least twice, it would be good to extract const as in original version

pdelewski (approved):

PR #753, opened 2024-09-11
Run E2E tests as a required check in merge queue

avelanarius:
Currently we don't run E2E tests on PRs or in merge queue, just on the main branch. That means that someone could merge a PR that ultimately fails on the main branch. Fix that by running E2E tests in the merge queue to catch the issue earlier.


avelanarius:
My rationale is that since I'm now changing the configuration, I nearly forgot to change E2E's configuration, which wouldn't have been caught before merge.

pdelewski (approved):

PR #757, opened 2024-09-12
Fix array handling

nablaone:


jakozaur (approved):
Lgtm, but it would be great if we add test too.

PR #760, opened 2024-09-12
Facet generalization and post-pancake cleanup

jakozaur:
Few items:
- We no longer got specialized SQL queries for facets, we use general logic
- Remove a lot of code related to pre-pancake world

Removed some tests which were too close too implementation that I removed.


nablaone (approved):
LGTM.

PR #765, opened 2024-09-13
Fix resolving indices when tables are auto-discovered

mieciu:
Users who haven't specified any table configurations aren't able to create data views today.

1. Because `_resolve` endpoint relies on schema registry, which - as turned out - has never been populated when the index configuration has been missing. 
2. Because `_field_caps` were **not** routed properly.
`matchedAgainstPattern` did rely only on config, not on schema registry, so it couldn't route queries to auto discovered tables. Therefore, timestamp field list has not been expanding and data view could not have been created.

This PR addresses both issues.

**How it works now:**
image

### DIFF
--- a/quesma/schema/registry.go
+++ b/quesma/schema/registry.go
@@ -39,6 +39,14 @@ type (
 func (s *schemaRegistry) loadSchemas() (map[TableName]Schema, error) {
 	definitions := s.dataSourceTableProvider.TableDefinitions()
 	schemas := make(map[TableName]Schema)
+	if len(*s.indexConfiguration) == 0 { // no index configs => table auto-discovery is enabled
+		for tableName, _ := range definitions {
+			fields := make(map[FieldName]Field)
+			existsInDataSource := s.populateSchemaFromTableDefinition(definitions, tableName, fields)
+			schemas[TableName(tableName)] = NewSchema(fields, existsInDataSource)
+		}
+		return schemas, nil
+	}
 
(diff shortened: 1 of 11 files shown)


quesma/schema/registry.go
```diff
@@ -39,6 +39,14 @@ type (
 func (s *schemaRegistry) loadSchemas() (map[TableName]Schema, error) {
 	definitions := s.dataSourceTableProvider.TableDefinitions()
 	schemas := make(map[TableName]Schema)
+	if len(*s.indexConfiguration) == 0 { // no index configs => table auto-discovery is enabled
```
avelanarius, line 42: In the near future, I plan to change the way auto discovery is enabled from "no index configs => auto discovery" to:
```yaml
# config of query processor:
  "*": # syntax not final
    "target": ["clickhouse"]
  ...
  "other_index": ...
```

That way you could mix autodiscovery and explicit index configs.

Maybe this is not necessary in this PR, but you could improve the logic to be able to merge `s.indexConfiguration` and definitions not present in `s.indexConfiguration`.

mieciu: When working on this further I noticed  that `len(*s.indexConfiguration) == 0` became a recurring pattern so I'll move it to dedicated function, which will also make the code more bulletproof

avelanarius: Nice!

avelanarius (approved):

PR #767, opened 2024-09-15
Schema field encoding additions

pdelewski:
This PR introduces changes for field encodings according to defined criteria by https://github.com/QuesmaOrg/quesma/blob/main/adr/5_nested_fields_representation.md

Fields are encoded during ingest with original name stored as comment 

image

### DIFF
--- a/quesma/ingest/parser.go
+++ b/quesma/ingest/parser.go
@@ -23,7 +23,18 @@ type CreateTableEntry struct {
 // Rendering columns to string
 func columnsToString(columnsFromJson []CreateTableEntry,
 	columnsFromSchema map[schema.FieldName]CreateTableEntry,
+	fieldEncodings map[schema.FieldEncodingKey]schema.EncodedFieldName,
+	tableName string,
 ) string {
+
+	reverseFieldEncoding := make(map[schema.EncodedFieldName]schema.FieldEncodingKey)
+
+	for k, v := range fieldEncodings {
+		if k.TableName == tableName {
+			reverseFieldEncoding[v] = k
+		}
+	}
+
 	var result strings.Builder
 	first := true
 	for _, columnFromJson := range columnsFromJson {
@@ -36,9 +47,9 @@ func columnsToString(columnsFromJson []CreateTableEntry,
 
 		if columnFromSchema, found := columnsFromSchema[schema.FieldName(columnFromJson.ClickHouseColumnName)]; found && !strings.Contains(columnFromJson.ClickHouseType, "Array") {
 			// Schema takes precedence over JSON (except for Arrays which are not currently handled)
-			result.WriteString(fmt.Sprintf("\"%s\" %s", columnFromSchema.ClickHouseColumnName, columnFromSchema.ClickHouseType))
+			result.WriteString(fmt.Sprintf("\"%s\" %s '%s'", columnFromSchema.ClickHouseColumnName, columnFromSchema.ClickHouseType+" COMMENT ", reverseFieldEncoding[schema.EncodedFieldName(columnFromSchema.ClickHouseColumnName)].FieldName))
 		} else {
-			result.WriteString(fmt.Sprintf("\"%s\" %s", columnFromJson.ClickHouseColumnName, columnFromJson.ClickHouseType))
+			result.WriteString(fmt.Sprintf("\"%s\" %s '%s'", columnFromJson.ClickHouseColumnName, columnFromJson.ClickHouseType+" COMMENT ", reverseFieldEncoding[schema.EncodedFieldName(columnFromJson.ClickHouseColumnName)].FieldName))
 		}
 
 		delete(columnsFromSchema, schema.FieldName(columnFromJson.ClickHouseColumnName))
@@ -52,7 +63,7 @@ func columnsToString(columnsFromJson []CreateTableEntry,
 			result.WriteString(",\n")
 		}
 		result.WriteString(util.Indent(1))
-		result.WriteString(fmt.Sprintf("\"%s\" %s", column.ClickHouseColumnName, column.ClickHouseType))
+		result.WriteString(fmt.Sprintf("\"%s\" %s '%s'", column.ClickHouseColumnName, column.ClickHouseType+" COMMENT ", reverseFieldEncoding[schema.EncodedFieldName(column.ClickHouseColumnName)].FieldName))
 	}
 	return result.String()
 }
@@ -114,10 +125,10 @@ func SchemaToColumns(schemaMapping *schema.Schema, nameFormatter TableColumNameF
 
 	for _, field := range schemaMapping.Fields {
 		var fType string
-
-		// We should never have dots in the field names, see 4 ADR
-		internalPropertyName := strings.Replace(field.InternalPropertyName.AsString(), ".", "::", -1)
-
+		// TODO we are using util.FieldToColumnEncoder directly here
+		// due to the fact that field encodings map is part of schema.Registry
+		// not schema.Schema and we don't have access to it here.
+		internalPropertyName := util.FieldToColumnEncoder(field.InternalPropertyName.AsString())
 		switch field.Type.Name {
 		default:
 			logger.Warn().Msgf("Unsupported field type '%s' for field '%s' when trying to create a table. Ignoring that field.", field.Type.Name, field.PropertyName.AsString())
--- a/quesma/ingest/processor.go
+++ b/quesma/ingest/processor.go
@@ -569,10 +569,6 @@ func generateSqlStatements(createTableCmd string, alterCmd []string, insert stri
 	return statements
 }
 
-func fieldToColumnEncoder(field string) string {
-	return strings.Replace(field, ".", "::", -1)
-}
-
 func (ip *IngestProcessor) processInsertQuery(ctx context.Context,
 	tableName string,
 	jsonData []types.JSON, transformer jsonprocessor.IngestTransformer,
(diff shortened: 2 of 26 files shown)


avelanarius:
Looking at failing tests in `registry_test.go`. 

Those tests fail because `internalToPublicFieldsEncodings` is empty and therefore `propertyName := FieldName(internalToPublicFieldsEncodings[column.Name])` all `propertyNames` get encoded into `""`

pdelewski:
> Looking at failing tests in `registry_test.go`.
> 
> Those tests fail because `internalToPublicFieldsEncodings` is empty and therefore `propertyName := FieldName(internalToPublicFieldsEncodings[column.Name])` all `propertyNames` get encoded into `""`

In this case I would expect to see the same name as public and internal one https://github.com/QuesmaOrg/quesma/pull/767/commits/75286bb2e4bc23ec38725d907df720bb702e3bd7

quesma/schema/registry.go
```diff
@@ -145,13 +185,17 @@ func (s *schemaRegistry) populateAliases(indexConfiguration config.IndexConfigur
...
 
-func (s *schemaRegistry) populateSchemaFromTableDefinition(definitions map[string]Table, indexName string, fields map[FieldName]Field) (existsInDataSource bool) {
+func (s *schemaRegistry) populateSchemaFromTableDefinition(definitions map[string]Table, indexName string, fields map[FieldName]Field, internalToPublicFieldsEncodings map[string]string) (existsInDataSource bool) {
 	tableDefinition, found := definitions[indexName]
 	if found {
 		logger.Debug().Msgf("loading schema for table %s", indexName)
 
 		for _, column := range tableDefinition.Columns {
-			propertyName := FieldName(strings.Replace(column.Name, "::", ".", -1))
+			propertyName := FieldName(internalToPublicFieldsEncodings[column.Name])
```
mieciu, line 194: ugh, that's really a publicFieldName, not a property, but it blends with that Field structur defined in `schema.go` which overall collides a little with the map of encodings

quesma/ingest/parser.go
```diff
@@ -114,10 +114,7 @@ func SchemaToColumns(schemaMapping *schema.Schema, nameFormatter TableColumNameF
 
 	for _, field := range schemaMapping.Fields {
 		var fType string
-
-		// We should never have dots in the field names, see 4 ADR
-		internalPropertyName := strings.Replace(field.InternalPropertyName.AsString(), ".", "::", -1)
-
+		internalPropertyName := util.FieldToColumnEncoder(field.InternalPropertyName.AsString())
```
mieciu, line 117: as chatted on call, we could've used `schemaMapping *schema.Schema` object instead of function here

quesma/ingest/processor.go
```diff
@@ -560,14 +564,30 @@ func (ip *IngestProcessor) processInsertQuery(ctx context.Context,
...
 	}
+	jsonData = processed
+
+	encodings := make(map[schema.FieldEncodingKey]string)
+	// we are doing two passes, e.g. calling transformFieldName twice
+	// first time we populate encodings map
+	// second time we do field encoding
+	// This is because we need to know all field encodings before we start
+	// transforming fields, otherwise we would mutate json and populating encodings
+	// which would introduce side effects
```
mieciu, line 575: yea that's a bummer, but I don't have any better idea for now. I think it might hit us faster than we might expect though, esp. that for clickhouse ingest its advices to have significantly larger bulk payloads than elasticsearch (hundreds of thousands vs few thousands)

quesma/ingest/processor.go
```diff
@@ -560,14 +564,30 @@ func (ip *IngestProcessor) processInsertQuery(ctx context.Context,
...
+	}
+	if ip.schemaRegistry != nil {
+		ip.schemaRegistry.UpdateFieldEncodings(encodings)
+	}
 	// Do field encoding here, once for all jsons
 	// This is in-place operation
 	for _, jsonValue := range jsonData {
 		transformFieldName(jsonValue, func(field string) string {
-			return fieldToColumnEncoder(field)
+			return util.FieldToColumnEncoder(field)
```
mieciu, line 588: Now when I think about those usages of `FieldToColumnEncoder`, perhaps this function should really part of schema registry? I could imagine registry being an interface or perhaps having it parametrized. But of course today it's not an option given he'd still use that to assemble the response :)

quesma/quesma/schema_transformer.go
```diff
@@ -534,6 +560,7 @@ func (s *SchemaCheckPass) Transform(queries []*model.Query) ([]*model.Query, err
 	}{
 		{TransformationName: "PhysicalFromExpressionTransformation", Transformation: s.applyPhysicalFromExpression},
 		{TransformationName: "WildcardExpansion", Transformation: s.applyWildcardExpansion},
+		{TransformationName: "FieldEncodingTransformation", Transformation: s.applyFieldEncoding},
```
mieciu, line 563: I'd comment why the position of `FieldEncodingTransformation` should be here :)

quesma/quesma/schema_transformer.go
```diff
@@ -526,6 +526,32 @@ func (s *SchemaCheckPass) applyTimestampField(indexSchema schema.Schema, query *
...
 
+func (s *SchemaCheckPass) applyFieldEncoding(indexSchema schema.Schema, query *model.Query) (*model.Query, error) {
+
+	visitor := model.NewBaseVisitor()
+
+	var err error
+
+	visitor.OverrideVisitColumnRef = func(b *model.BaseExprVisitor, e model.ColumnRef) interface{} {
+		if _, ok := indexSchema.ResolveField(e.ColumnName); ok {
+			return model.NewColumnRef(util.FieldToColumnEncoder(e.ColumnName))
```
mieciu, line 537: same here we could use `indexSchema schema.Schema` struct

quesma/util/utils.go
```diff
@@ -852,6 +852,9 @@ func replaceNonAlphabetic(str string) string {
 }
 
 func FieldToColumnEncoder(field string) string {
+	if len(field) == 0 {
```
mieciu, line 855: This function definitely needs a comment like
```
FieldToColumnEncoder takes publicly-facing Elasticsearch field name and using well-defined algorithm converts it to "legit" column name in ClickHouse. This encoding needs to happen because we don't want to have dots, colons or any other non-alphanumerical characters in columns names.  
```

quesma/util/utils.go
```diff
@@ -852,6 +852,9 @@ func replaceNonAlphabetic(str string) string {
...
+		return field
+	}
```
mieciu, line 857: I think this `FieldToColumnEncoder` deserves just a few, very simple unit tests just in case anyone accidentally thinks that "oh, let's change this" after few months from now 🙃 
Also, a table style test would nicely illustrate how the algorithm and expected output looks like in general

quesma/schema/registry.go
```diff
@@ -5,21 +5,29 @@ package schema
...
 )
 
 type (
 	Registry interface {
 		AllSchemas() map[TableName]Schema
 		FindSchema(name TableName) (Schema, bool)
 		UpdateDynamicConfiguration(name TableName, table Table)
+		UpdateFieldEncodings(encodings map[FieldEncodingKey]string)
+	}
+	FieldEncodingKey struct {
```
mieciu, line 19: as chatted on Slack, I think what really intricate with this structure at this moment is that I was trying to come up with a better naming - because we should be super precise which is public and which is internal fieldname. At this point we should ditch the generic "fieldName" I think and be super specific about what kind of field name it is specifically - public-facing or internal.

Then I realized it may be better for this encodings to actually live in the `Schema` object where we already "kind-of" keep this information? `schema.go`:
image (16)


If that brings too much hassle within this already huge PR, I'm good with keeping this duplication, but I'd advice putting a big fat warning saying "TODO HEY THIS IS SOMETHING WE REALLY NEED TO CLEANUP" 😉

quesma/schema/registry.go
```diff
@@ -5,21 +5,29 @@ package schema
...
+		FieldName string
 	}
 	schemaRegistry struct {
 		// index configuration overrides always take precedence
 		indexConfiguration      *map[string]config.IndexConfiguration
 		dataSourceTableProvider TableProvider
 		dataSourceTypeAdapter   typeAdapter
 		dynamicConfiguration    map[string]Table
+		fieldEncodingsLock      sync.RWMutex
+		fieldEncodings          map[FieldEncodingKey]string
```
mieciu, line 30: additionally to what I've wrote about this belonging to `Schema` and removing the redundant reference to `table name` - I think we should make this `map[FieldEncodingKey]InternalFieldName` or maybe `map[FieldEncodingKey]EncodedFieldName` so just adding a meaningful type alias :)

mieciu (changes requested):
That's a great work overall !!! 

I just left few thoughts, esp the duplicated information about table in `map[FieldEncodingKey]string` and being more specific with names (which are are public vs private fields makes this code a little more elusive than it should be. But if you think that's better to handle separately - I'm really open to merge this as is and just live with this - already way better - codebase 💪

quesma/ingest/parser.go
pdelewski: For now `schema.Schema` does not store info about field encodings. They are stored in `schema.Registry`. I will add comment here, but it would be cleaner and safer to do that in subsequent PR

quesma/schema/registry.go
pdelewski: yes, we can try to fix that later

quesma/ingest/processor.go
pdelewski: true, but this can be further optimized. One of the method would be to rewrite `transformFieldName` function, but I think it's too much for this PR

quesma/ingest/processor.go
pdelewski: Maybe, for it's just utility function, but it could be moved or invoked from `schema.Registry`

quesma/quesma/schema_transformer.go
pdelewski: Will add comments

quesma/util/utils.go
pdelewski: Tests were added in previous PR https://github.com/QuesmaOrg/quesma/blob/main/quesma/util/utils_test.go#L825

quesma/quesma/schema_transformer.go
pdelewski: Added comment

quesma/schema/registry.go
pdelewski: I changed signature `UpdateFieldEncodings(encodings map[FieldEncodingKey]EncodedFieldName)` and refined type and also add a comment for further improvements

pdelewski:
> That's a great work overall !!!
> 
> I just left few thoughts, esp the duplicated information about table in `map[FieldEncodingKey]string` and being more specific with names (which are are public vs private fields makes this code a little more elusive than it should be. But if you think that's better to handle separately - I'm really open to merge this as is and just live with this - already way better - codebase 💪

I made some improvements, now we have `map[FieldEncodingKey]EncodedFieldName`

mieciu (approved):
:shipit: !

PR #768, opened 2024-09-16
Resolve table name with schema

mieciu:


quesma/clickhouse/quesma_communicator.go
```diff
@@ -160,7 +160,6 @@ func executeQuery(ctx context.Context, lm *LogManager, query *model.Query, field
 	//
 
 	settings := make(clickhouse.Settings)
-	settings["readonly"] = "1"
```
nablaone, line 163: Why is that?

nablaone (approved):
LGTM.

quesma/clickhouse/quesma_communicator.go
mieciu: Apparently that causes troubles with Hydrolix queries - I'll link you some stack traces.

pdelewski, line 163: Please add comment

pdelewski (approved):

PR #770, opened 2024-09-16
Array types - hybrid solution

nablaone:
This PR adds another way to handle queries with array columns. If we have an aggregation we can expand (arrayJoin) the array. Computed results looks better. 

Here are examples of how it looks:

Screenshot 2024-09-16 at 17 42 56

Screenshot 2024-09-16 at 17 42 32

### DIFF
--- a/quesma/quesma/schema_transformer.go
+++ b/quesma/quesma/schema_transformer.go
@@ -4,6 +4,7 @@ package quesma
 
 import (
 	"fmt"
+	"quesma/common_table"
 	"quesma/logger"
 	"quesma/model"
 	"quesma/model/typical_queries"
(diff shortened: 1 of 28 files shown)


jakozaur:
Great direction, I got some idea how to solve it for good. Though it needs to work with window functions:
https://fiddle.clickhouse.com/082ff22e-b853-45ef-8d96-05f54131ad39

jakozaur:
I debate, but it would probably still make a lot of sense to merge this in its current form.

quesma/quesma/schema_array_transformer.go
```diff
@@ -100,3 +100,87 @@ func NewArrayTypeVisitor(resolver arrayTypeResolver) model.ExprVisitor {
 	}
 	return visitor
 }
+
+func checkIfGoupingByArrayColumn(stm model.SelectCommand, resolver arrayTypeResolver) bool {
```
trzysiek, line 104: what `stm` stands for?

trzysiek, line 104:
```suggestion
func checkIfGroupingByArrayColumn(stm model.SelectCommand, resolver arrayTypeResolver) bool {
```

quesma/quesma/schema_transformer.go
```diff
@@ -293,7 +293,13 @@ func (s *SchemaCheckPass) applyArrayTransformations(indexSchema schema.Schema, q
 		return query, nil
 	}
 
-	visitor := NewArrayTypeVisitor(arrayTypeResolver)
+	var visitor model.ExprVisitor
+
+	if checkIfGoupingByArrayColumn(query.SelectCommand, arrayTypeResolver) {
```
trzysiek, line 298:
```suggestion
	if checkIfGroupingByArrayColumn(query.SelectCommand, arrayTypeResolver) {
```

quesma/quesma/schema_array_transformer.go
nablaone: `stm` == `statement`

I'll change to `selectCommand`.

trzysiek (approved):

PR #771, opened 2024-09-17
Remove "Disabled" from index conf, switch code to use QueryTarget, IngestTarget

avelanarius:
This commit removes the `disabled` option from index configuration, as it's now possible to reflect it via target configuration:

Previously:

```yaml
my_index:
  disabled: true
```

Now:

```yaml
my_index:
  target: [ "my-elastic-connector" ]
```

This highlights a problem with `disabled` - it meant that the index should be read/written from Elastic, but a user might have incorrectly thought that `disabled` means that all queries to the index will get rejected.

Upon removing `disabled`, all internal Quesma code is now modified to use the `QueryTarget` and `IngestTarget` arrays directly, paving the way for more flexible target configuration in the future (next PRs).

### DIFF
--- a/quesma/quesma/router_test.go
+++ b/quesma/quesma/router_test.go
@@ -42,7 +42,7 @@ func Test_matchedAgainstConfig(t *testing.T) {
 
 			req := &mux.Request{Params: map[string]string{"index": tt.index}, Body: tt.body}
 
-			assert.Equalf(t, tt.want, matchedExact(&tt.config).Matches(req), "matchedAgainstConfig(%v), index: %s", tt.config, tt.index)
+			assert.Equalf(t, tt.want, matchedExactQueryPath(&tt.config).Matches(req), "matchedExactQueryPath(%v), index: %s", tt.config, tt.index)
 		})
 	}
 }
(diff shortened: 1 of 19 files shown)


quesma/quesma/router_test.go
```diff
@@ -137,8 +137,14 @@ func Test_matchedAgainstPattern(t *testing.T) {
 	}
 }
 
-func indexConfig(name string, disabled bool) config.QuesmaConfiguration {
-	return config.QuesmaConfiguration{IndexConfig: map[string]config.IndexConfiguration{name: {Name: name, Disabled: disabled}}}
+func indexConfig(name string, elastic bool) config.QuesmaConfiguration {
```
mieciu, line 140:
```suggestion
func indexConfig(name string, targetsElastic bool) config.QuesmaConfiguration {
```

mieciu (commented):
In general LGTM, but I noticed repetitive use of 
`slices.Contains(indexConfig.QueryTarget, config.ClickhouseTarget)`
which made me think that perhaps we should extract this (and similar function) as a method of `indexConfig`? WDYT?

mieciu (approved):
Giving approve to procedurally unblock you, as we've agreed on slack that https://github.com/QuesmaOrg/quesma/pull/771#pullrequestreview-2312709770 is going to be addressed 😉

PR #773, opened 2024-09-17
Always fill-in InternalPropertyType in populateSchemaFromTableDefinition

avelanarius:
Previously `InternalPropertyType` field was only filled-in in only one code path in `populateSchemaFromTableDefinition`. However, it should be filled-in in all cases - table definition is the only source of this information and `populateSchemaFromTableDefinition` is the only place it's filled-in - if it's not correctly set there, it will stay empty forever.

This fixes a problem where "PUT mapping" first filled out the schema registry and `populateSchemaFromTableDefinition` did not update `InternalPropertyType`, leaving it empty, which caused failures on several eCommerce panels (`schema_array_transformer.go` did not get correct `InternalPropertyType`, so it did not perform any required transformation).


nablaone (approved):
LGTM

PR #777, opened 2024-09-19
Common table - query multiple indexes 

nablaone:
This PR adds support for querying multiple indexes.  It's limited to indexes stored in `common table`.

Screenshot 2024-09-19 at 12 37 33

Screenshot 2024-09-19 at 12 37 53

There are a few things to do:
- check why some columns are not translated properly (see `handleDottedTColumnNames` transformer)
- `table.Clickhouse` removal from `ClickhouseQueryTranslator`
- testing, more testing

### DIFF
--- a/quesma/model/query.go
+++ b/quesma/model/query.go
@@ -3,6 +3,7 @@
 package model
 
 import (
+	"quesma/schema"
 	"time"
 )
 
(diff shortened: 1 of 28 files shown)


quesma/model/query.go
```diff
@@ -47,7 +48,12 @@ type (
 		TransformationHistory TransformationHistory // it can be optional
 
 		Type      QueryType
-		TableName string
+		TableName string // TODO delete this and use Indexes instead
```
jakozaur, line 51: Maybe mark as deprecated.

jakozaur (approved):
Looks good, though wish we have some tests or that.

PR #779, opened 2024-09-20
Runtime mappings

nablaone:
This PR adds basic (rudimentary) support for runtime mappings. 
- it's limited only to mappings that are passed  in a query
- it supports only one expression now, fixes `[Logs] Web Traffic`  dashboard

Screenshot 2024-09-30 at 16 36 23

Screenshot 2024-09-30 at 16 36 42


quesma/queryparser/runtime_mappings.go
```diff
@@ -0,0 +1,54 @@
...
+	"quesma/model"
+	"quesma/quesma/types"
+)
+
+func ParseRuntimeMappings(body types.JSON) map[string]model.RuntimeMapping {
+
+	result := make(map[string]model.RuntimeMapping)
+
+	if runtimeMappings, ok := body["runtime_mappings"]; ok {
+		if runtimeMappingsMap, ok := runtimeMappings.(map[string]interface{}); ok {
```
jakozaur, line 15: In other places we use `JsonMap` instead of `map[string]interface{}`.

jakozaur: PErhaps logging error or warn, when type is different would be also nice.

jakozaur (approved):
Can be better, but right step.

PR #780, opened 2024-09-22
Percentile combinator

trzysiek:


quesma/queryparser/pancake_sql_query_generation.go
```diff
@@ -53,6 +53,12 @@ func (p *pancakeSqlQueryGenerator) generateAccumAggrFunctions(origExpr model.Exp
 			// This is ClickHouse specific: https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators
 			return model.NewFunction(origFunc.Name+"State", origFunc.Args...), origFunc.Name + "Merge", nil
 		}
+
+		// We use sumArray here, because "quantiles" returns an array.
+		// We probably could simplify the code a bit to use "quantile" instead, and just "sum" here.
+		if strings.HasPrefix(origFunc.Name, "quantiles") {
+			return origFunc, "sumArray", nil
+		}
```
jakozaur, line 61: Sum of quantiles does not seem correct. I would add quantiles to same condition that creates State and Merge, but likely you need to do some parsing.

jakozaur (changes requested):
Thanks for fixing, but likely we need to use State and Merge.

jakozaur (approved):

PR #781, opened 2024-09-22
Very small code cleanup

trzysiek:
Just a small removal of some pre-pancake code (+ fix expected results for 1 test)


quesma/testdata/aggregation_requests.go
trzysiek, line 1: This test had incorrect expected column results (column names didn't match the expected SQL)

jakozaur (approved):

PR #782, opened 2024-09-23
Dual write support in index config

avelanarius:
This PR makes it possible to configure dual writes by specifying two targets in ingest processor:

```yaml
my_index:
  target: [ my-clickhouse-data-source, my-minimal-elasticsearch ]
```

The limitation, that the ingest processor and query processor must have the same configuration, is now lifted (so you can specify reading from Elastic, but ingesting into both ClickHouse and Elastic).

The implementation of `/_doc` endpoint is now simplified to use the existing logic of bulk ingest (which already had a proper support for dual writes).

### DIFF
--- a/quesma/quesma/config/config_v2.go
+++ b/quesma/quesma/config/config_v2.go
@@ -245,6 +245,11 @@ func (c *QuesmaNewConfiguration) validatePipelines() error {
 		if ingestProcessor.Type != QuesmaV1ProcessorIngest && ingestProcessor.Type != QuesmaV1ProcessorNoOp {
 			return fmt.Errorf("ingest pipeline must have ingest-type or noop processor")
 		}
+		for _, indexConf := range ingestProcessor.Config.IndexConfig {
+			if len(indexConf.Optimizers) != 0 {
+				return fmt.Errorf("configuration of index '%s' in '%s' processor cannot have any optimizers, this is only a feature of query processor", ingestPipeline.Processors[0], indexConf.Name)
+			}
+		}
 		queryProcessor := c.getProcessorByName(queryPipeline.Processors[0])
 		if queryProcessor == nil {
 			return fmt.Errorf(fmt.Sprintf("query processor named [%s] not found in configuration", ingestPipeline.Processors[0]))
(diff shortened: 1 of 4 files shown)


quesma/quesma/functionality/doc/doc.go
```diff
@@ -4,29 +4,18 @@ package doc
...
-			tableName = cfg.IndexConfig[tableName].Override
-		}
-		nameFormatter := clickhouse.DefaultColumnNameFormatter()
-		transformer := jsonprocessor.IngestTransformerFor(tableName, cfg)
-		return ip.ProcessInsertQuery(ctx, tableName, types.NDJSON{body}, transformer, nameFormatter)
-	})
+func Write(ctx context.Context, tableName *string, body types.JSON, ip *ingest.IngestProcessor, cfg *config.QuesmaConfiguration, phoneHomeAgent telemetry.PhoneHomeAgent) (bulk.BulkItem, error) {
+	// Translate single doc write to a bulk request, reusing exiting logic of bulk ingest
+	results, err := bulk.Write(ctx, tableName, []types.JSON{
+		map[string]interface{}{"index": map[string]interface{}{"_index": *tableName}},
```
nablaone, line 17: This is magic. Maybe we should extract a function here.

quesma/quesma/config/config_v2.go
```diff
@@ -245,6 +245,11 @@ func (c *QuesmaNewConfiguration) validatePipelines() error {
 		if ingestProcessor.Type != QuesmaV1ProcessorIngest && ingestProcessor.Type != QuesmaV1ProcessorNoOp {
 			return fmt.Errorf("ingest pipeline must have ingest-type or noop processor")
 		}
+		for _, indexConf := range ingestProcessor.Config.IndexConfig {
+			if len(indexConf.Optimizers) != 0 {
+				return fmt.Errorf("configuration of index '%s' in '%s' processor cannot have any optimizers, this is only a feature of query processor", ingestPipeline.Processors[0], indexConf.Name)
```
nablaone, line 250: This can be a warning. I'm not sure about the policy (strictness) regarding our configuration.

quesma/quesma/router.go
```diff
@@ -418,24 +419,12 @@ func resolveIndexResult(sources elasticsearch.Sources) (*mux.Result, error) {
 		StatusCode: http.StatusOK}, nil
 }
 
-func indexDocResult(index string, statusCode int) (*mux.Result, error) {
-	body, err := json.Marshal(indexDocResponse{
```
nablaone, line 422: I don't understand why we return something different here.

mieciu (approved):
:shipit:

PR #784, opened 2024-09-23
Introduce auth 

mieciu:
This PR introduces an authentication for Elasticsearch API exposed by Quesma, which - to put it short - defers to Elasticsearch when authenticating incoming requests. 
For more implementation details, there's an [ADR](https://github.com/QuesmaOrg/quesma/blob/a46257bc70567016a4029837f28ba29ea32c03f0/adr/6_relying_on_elasticsearch_for_auth.MD).

### Notes for reviewer
1. I am adding a `disableAuth: true` option to Quesma frontend connector configuration so that our local dev setups maintain their lean nature 😉 
2. I am also adding `elasticsearch.NewSimpleClient` as I have noticed that creating a dumb `http.Client` which shoots ES API request occurs in few places (`NewElasticJSONDatabase`, `NewIndexManagement`), so we can reuse that logic. However, I'd prefer doing that refactor in a separate PR, so for now only auth middleware is using it.

### DIFF
--- a/quesma/elasticsearch/client.go
+++ b/quesma/elasticsearch/client.go
@@ -1,3 +1,5 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
 package elasticsearch
 
 import (
--- a/quesma/elasticsearch/client.go
+++ b/quesma/elasticsearch/client.go
@@ -4,10 +4,12 @@ package elasticsearch
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"quesma/logger"
 	"quesma/quesma/config"
 	"time"
 )
@@ -31,18 +33,18 @@ func NewSimpleClient(configuration *config.ElasticsearchConfiguration) *SimpleCl
 		config: configuration,
 	}
 }
-func (es *SimpleClient) Request(method, endpoint string, payload interface{}) (*http.Response, error) {
+func (es *SimpleClient) Request(ctx context.Context, method, endpoint string, payload interface{}) (*http.Response, error) {
 	body, err := json.Marshal(payload)
 	if err != nil {
 		return nil, err
 	}
-	return es.doRequest(method, endpoint, body, nil)
+	return es.doRequest(ctx, method, endpoint, body, nil)
 }
 
-func (es *SimpleClient) Authenticate(authHeader string) bool {
-	resp, err := es.doRequest("GET", "_security/_authenticate", nil, http.Header{"Authorization": {authHeader}})
+func (es *SimpleClient) Authenticate(ctx context.Context, authHeader string) bool {
+	resp, err := es.doRequest(ctx, "GET", "_security/_authenticate", nil, http.Header{"Authorization": {authHeader}})
 	if err != nil {
-		fmt.Println("Error sending request:", err)
+		logger.ErrorWithCtx(ctx).Msgf("error sending request: %v", err)
 		return false
 	}
 	defer resp.Body.Close()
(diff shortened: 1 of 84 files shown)


quesma/elasticsearch/client.go
```diff
@@ -0,0 +1,63 @@
...
+	"quesma/quesma/config"
+)
+
+type SimpleClient struct {
+	client *http.Client
+	config *config.ElasticsearchConfiguration
+}
+
+func NewSimpleClient(configuration *config.ElasticsearchConfiguration) *SimpleClient {
+	client := &http.Client{
```
nablaone, line 20: We should specify timeouts here.

quesma/quesma/auth_middleware.go
```diff
@@ -0,0 +1,69 @@
...
+	if auth == "" {
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return
+	}
+	if _, ok := a.authHeaderCache.Load(auth); ok {
+		a.nextHttpHandler.ServeHTTP(w, r)
+		return
+	}
+
+	if authenticated := a.esClient.Authenticate(auth); authenticated {
```
nablaone, line 45: Here we should pass `context.Context` of the current request to the http client.

quesma/quesma/auth_middleware.go
```diff
@@ -0,0 +1,69 @@
...
+		a.authHeaderCache.Store(auth, struct{}{})
```
nablaone, line 46: Why we don't cache `false` results?

quesma/quesma/auth_middleware.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func NewAuthMiddleware(next http.Handler, esConf config.ElasticsearchConfiguration) http.Handler {
+	esClient := elasticsearch.NewSimpleClient(&esConf)
+	middleware := &authMiddleware{nextHttpHandler: next, esClient: *esClient, cacheWipeInterval: cacheWipeInterval}
+	go middleware.startCacheWipeScheduler()
+	return middleware
+}
+
+func (a *authMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
nablaone, line 34: This middleware is quiet. That's fine for production deployment.  It will be a PITA to solve any auth issues on the test/stg environment.  Maybe we should have a flag and log every unsuccessful attempt.

nablaone (changes requested):
We should handle panics.

nablaone:
This middleware can be applied to the UI endpoint either.

quesma/quesma/auth_middleware.go
mieciu: At first I was leaning towards more complex cache: storing both positive and negative attempts plus time of adding an item so that we wipe old entries based on that.

Then I had some afterthoughts and chosen this simpler implementation to save on PRs LoC :) 

Main purpose of this cache was to make Quesma's responses faster - even if someone will be hammering Elasticsearch with failed auth attempts - these will still be 401's anyways so I didn't see much of a value from caching those :)

quesma/quesma/auth_middleware.go
```diff
@@ -0,0 +1,69 @@
...
+		a.authHeaderCache.Store(auth, struct{}{})
+	} else {
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return
+	}
+
+	a.nextHttpHandler.ServeHTTP(w, r)
+}
+
+func (a *authMiddleware) startCacheWipeScheduler() {
```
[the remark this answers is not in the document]
mieciu: https://github.com/QuesmaOrg/quesma/pull/784/commits/26f5c5901952c684bc89fca60394b6acb8141bd0

quesma/quesma/auth_middleware.go
mieciu: I was actually thinking about bringing auth-related messaging to debug logs, WDYT?

quesma/quesma/auth_middleware.go
mieciu: https://github.com/QuesmaOrg/quesma/pull/784/commits/0faaa71db1db89ea3dfed17c86ebbabc29bd8aab

quesma/quesma/auth_middleware.go
nablaone: I don't know how elastic works. What happens when we're hammering with failed auths? Does Elastic locks that account? If yes, these requests are ours, even if we're proxy them.

quesma/quesma/auth_middleware.go
nablaone: That's fine. I think we can run in debug mode in evaluation scenario.

nablaone (approved):
LGTM

quesma/quesma/auth_middleware.go
mieciu: Nothing like that happens as far as I'm concerned within Elasticsearch.
External auth providers (e.g. SAML) might do such things, but that's a different thing :)

PR #788, opened 2024-09-23
`Order by: some_other_aggregation` enhancements

trzysiek:
A couple of small enhancements here:
* During Kibana QA, I found out that it really happens a lot that in `order by: some_other_aggregation`, this aggregation is not a direct child, but some further descender in the aggregation tree, so added support for it (before `order by: "agg1"` worked, now `order by: "agg1>agg2>agg3"` also works (`agg2` is a child of `agg1`, `agg3` of `agg2`, etc.)
* Before we had support for
a) `order by: "2"` (where `2` is a metric aggr with single value like e.g. `avg`)
b) `order by: "2.10"` (where `2` is a `percentile[s|_ranks]` aggr, and `10` is a percentile)
Here added also support for `2.count`, `2.std_deviation`, and other stats from `stats` or `extended_stats` aggregations. 
* Before we could only order by 1 expression, 2 or more weren't supported. Fixed that.

Sorry for quite a big PR, but it turned out to be like that out of necessity. E.g. I implemented the last point only because without it, proper tests for previous points would need to be much larger 😆


jakozaur (commented):
I really like the change, but likely need code walkthrough. Still wonder whether we should do this in different place.

jakozaur (approved):

PR #789, opened 2024-09-24
Dual read configuration, default index configuration, disabling ingest for an index

avelanarius:
This PR is dependent on #782. (first commit is from that PR)

This PR improves the index configuration in several ways:
- Dual read (A/B testing) can be enabled by specifying two targets in query configuration of index:
  ```
  my_index:
    target: [clickhouse, elastic]
  ```
  The first target is the "primary" target - the returned result from Quesma will be from that target. Currently you can only specify Clickhouse as the primary target (due to A/B testing limitations).
- Add mandatory default index configuration `*`. Before this improvement, a user couldn't easily specify the behavior of Quesma for non-configured indexes. Now the user can specify the configuration of `*` index name, which will specify the default/fallback behavior. For example, specifying this in the query configuration:
   ```
   "*":
      target: [clickhouse]
   ```
   will enable autodiscovery.
- Make it possible to disable ingest for an index by specifying empty target list (`target: []`)

### DIFF
--- a/quesma/quesma/config/config_v2.go
+++ b/quesma/quesma/config/config_v2.go
@@ -333,8 +333,8 @@ func (c *QuesmaNewConfiguration) validateProcessor(p Processor) error {
 	if p.Type == QuesmaV1ProcessorQuery || p.Type == QuesmaV1ProcessorIngest {
 		for indexName, indexConfig := range p.Config.IndexConfig {
 			if p.Type == QuesmaV1ProcessorQuery {
-				if len(indexConfig.Target) != 1 {
-					return fmt.Errorf("configuration of index %s must have exactly one target (query processor)", indexName)
+				if len(indexConfig.Target) != 1 && len(indexConfig.Target) != 2 {
+					return fmt.Errorf("configuration of index %s must have one or two targets (query processor)", indexName)
 				}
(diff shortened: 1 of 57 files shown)


mieciu (approved):

quesma/quesma/config/config_v2.go
```diff
@@ -508,6 +519,19 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 				processedConfig.QueryTarget = append(processedConfig.QueryTarget, ClickhouseTarget)
 			}
 
+			if len(indexConfig.QueryTarget) == 2 && !(indexConfig.QueryTarget[0] == ClickhouseTarget && indexConfig.QueryTarget[1] == ElasticsearchTarget) {
+				errAcc = multierror.Append(errAcc, fmt.Errorf("index %s has invalid dual query target configuration - when you specify two targets, Elastic has to be the primary one and ClickHouse has to be the secondary one", indexName))
```
trzysiek, line 523: Error message doesn't match the code, I think? In `if` `[0]` is Clickhouse, `[1]` Elastic, not the other way around

trzysiek (approved):

PR #798, opened 2024-09-25
Database Column Comments as Metadata Storage – Part 2

nablaone:
This PR changes the way we store metadata in column comments. Now we store it as a structure. 

Screenshot 2024-09-25 at 14 52 59

### DIFF
--- a/quesma/comment_metadata/comment_metadata.go
+++ b/quesma/comment_metadata/comment_metadata.go
@@ -40,7 +40,7 @@ func UnmarshallCommentMetadata(s string) (*CommentMetadata, error) {
 	groups := rx.FindStringSubmatch(s)
 
 	if len(groups) == 0 {
-		return nil, fmt.Errorf("quesma metadata not found")
+		return nil, nil // no metadata, we return nil here, that's not an error
 	}
 
 	if len(groups) != 3 {
--- a/quesma/comment_metadata/comment_metadata.go
+++ b/quesma/comment_metadata/comment_metadata.go
@@ -1,0 +1,76 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+package comment_metadata
+
+import (
+	"fmt"
+	"net/url"
+	"regexp"
+)
+
+const ElasticFieldName = "fieldName"
+
+const commentMetadataVersion = "1"
+const metadataPrefix = "quesmaMetadata:"
+const versionParameter = "v"
+
+type CommentMetadata struct {
+	Values map[string]string
+}
+
+func NewCommentMetadata() *CommentMetadata {
+	return &CommentMetadata{
+		Values: make(map[string]string),
+	}
+}
+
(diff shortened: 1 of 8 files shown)


quesma/comment_metadata/comment_metadata.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"regexp"
+)
+
+const ElasticFieldName = "fieldName"
+
+const commentMetadataVersion = "1"
```
mieciu, line 13:
```suggestion
const metadataVersion = "1"
```
I think in this place `comment` is a little confusing

quesma/comment_metadata/comment_metadata.go
```diff
@@ -0,0 +1,76 @@
...
+const metadataPrefix = "quesmaMetadata:"
+const versionParameter = "v"
```
mieciu, line 15: I think  we could ditch this additional const `versionParameter` and just use `metadataVersion` and let it be `v1`  for simplicity. parsing will become more clear

mieciu (changes requested):
I think metadata version should be part of the metadata prefix instead of being a special metadata item

quesma/schema/registry.go
```diff
@@ -201,7 +202,15 @@ func (s *schemaRegistry) populateSchemaFromTableDefinition(definitions map[strin
 				// if field encodings are not coming from ingest e.g. encodings map
 				// is empty, read them from persistent storage, e.g. column comment
 			} else if len(column.Comment) > 0 {
-				propertyName = FieldName(column.Comment)
+				propertyName = FieldName(column.Name)
+				metadata, err := comment_metadata.UnmarshallCommentMetadata(column.Comment)
+				if err != nil {
+					logger.Warn().Msgf("error unmarshalling column '%s' (table: %s)  comment metadata: %s %v", indexName, column.Name, column.Comment, err)
+				} else {
+					if fieldName, ok := metadata.Values[comment_metadata.ElasticFieldName]; ok {
```
mieciu, line 210: I know this becomes a little if-ology but do if unmarshalling went fine and `metadata.Values` doesn't contain  `comment_metadata.ElasticFieldName` should we log this as something **really** strange ?

to reduce if's maybe you could use
```go
if metadata, err := comment_metadata.UnmarshallCommentMetadata(column.Comment); err == nil {
```

mieciu: Now that I think about it - perhaps we should provide a little facade method and let this logic live in
`metadata.getElasticsearchFieldName` ? 🤔 
Just a suggestion, not really a request.

nablaone: We don't keep in a schema any issues occurred while reading the table definition. In this case, if we cannot parse the metadata, we will expose that column with the default name. This can be confusing for the end user. 

We should log a warning if the format is invalid only. I'll change it.

mieciu (commented):
So it's `quesmaMetadataV1:...`now, nice!

mieciu (approved):
👍

PR #805, opened 2024-09-26
Common table - missing test - part#1  ingest

nablaone:
This PR 
1. Add tests.
2. Fix bugs found by these tests.


quesma/ingest/processor.go
```diff
@@ -882,12 +908,12 @@ func (ip *IngestProcessor) Ping() error {
 	return ip.chDb.Ping()
 }
 
-func NewEmptyIngestProcessor(cfg *config.QuesmaConfiguration, chDb *sql.DB, phoneHomeAgent telemetry.PhoneHomeAgent, loader chLib.TableDiscovery, schemaRegistry schema.Registry, virtualTableStorage persistence.JSONDatabase) *IngestProcessor {
+func NewIngestProcessor(cfg *config.QuesmaConfiguration, chDb *sql.DB, phoneHomeAgent telemetry.PhoneHomeAgent, loader chLib.TableDiscovery, schemaRegistry schema.Registry, virtualTableStorage persistence.JSONDatabase) *IngestProcessor {
```
mieciu, line 911: `NewIngestProcessor` is the only constructor that should live in this file - rest of them are for test purposes only, so IMO let's move them to the test code.
I'd suggest also have them calling `NewIngestProcessor` instead of initializing struct every time.

quesma/ingest/processor.go
```diff
@@ -882,12 +908,12 @@ func (ip *IngestProcessor) Ping() error {
...
 	ctx, cancel := context.WithCancel(context.Background())
 	return &IngestProcessor{ctx: ctx, cancel: cancel, chDb: chDb, tableDiscovery: loader, cfg: cfg, phoneHomeAgent: phoneHomeAgent, schemaRegistry: schemaRegistry, virtualTableStorage: virtualTableStorage}
 }
 
-func NewIngestProcessor(tables *TableMap, cfg *config.QuesmaConfiguration) *IngestProcessor {
+func NewIngestProcessorTableMapConfigEmpty(tables *TableMap, cfg *config.QuesmaConfiguration) *IngestProcessor {
```
mieciu, line 916:
```suggestion
func NewIngestProcessorWithEmptyTableMap(tables *TableMap, cfg *config.QuesmaConfiguration) *IngestProcessor {
```

quesma/ingest/processor.go
```diff
@@ -834,11 +843,28 @@ func (ip *IngestProcessor) storeVirtualTable(table *chLib.Table) error {
...
+		// We don't want to store attributes columns in the virtual table
+		if col.Name == chLib.AttributesValuesColumn || col.Name == chLib.AttributesMetadataColumn {
+			continue
+		}
+		columnsToStore = append(columnsToStore, col.Name)
+	}
+
+	// We always want to store timestamp in the virtual table
+	// if it's not already there
+	if slices.Index(columnsToStore, model.TimestampFieldName) == -1 {
```
mieciu, line 857:
```suggestion
	if slices.Contains(columnsToStore, model.TimestampFieldName) {
```
Could we use this here instead?

trzysiek:
Some general styling comment in tests, I personally prefer to use stuff like `assert.True(t, ok)` instead of 
```
if !ok {
    t.Fatal("expected ok")    
}
```
Shorted code and personally easier to read as well. I don't mind verbose ones, but maybe you'll find mine nicer too.

jakozaur:
I like change, but would second @trzysiek . We should use `assert` as much as possible.

jakozaur (approved):

PR #811, opened 2024-09-27
Fix missing a metadata comment

nablaone:
Was:

Screenshot 2024-09-27 at 11 38 12

is:
Screenshot 2024-09-27 at 11 29 30

### DIFF
--- a/quesma/ingest/insert_test.go
+++ b/quesma/ingest/insert_test.go
@@ -359,6 +359,7 @@ func dateTime(name string) *clickhouse.Column {
 	}
 }
 
+// TestCreateTableIfSomeFieldsExistsInSchemaAlready verifies the result `CREATE TABLE` statement when the ingested JSON doesn't contain *all* the columns from the table. In this case we should take the column data from schema and make sure the final statement includes all the columns. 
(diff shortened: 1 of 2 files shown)


quesma/ingest/insert_test.go
```diff
@@ -356,3 +358,81 @@ func dateTime(name string) *clickhouse.Column {
 		Modifiers: "CODEC(DoubleDelta, LZ4)",
 	}
 }
+
+func TestCreateTableIfSomeFieldsExistsInSchemaAlready(t *testing.T) {
```
mieciu, line 362:
```suggestion
// TestCreateTableIfSomeFieldsExistsInSchemaAlready verifies the result `CREATE TABLE` statement when the ingested JSON doesn't contain *all* the columns from the table. In this case we should take the column data from schema and make sure the final statement includes all the columns. 
func TestCreateTableIfSomeFieldsExistsInSchemaAlready(t *testing.T) {
```

mieciu: this is quite complex case, I think this comment will explain it better than any name :)

mieciu (approved):

PR #793, opened 2024-09-24
Fix total hits count

trzysiek:
edit: fix for https://github.com/QuesmaOrg/quesma/issues/972

Before us: (17x bigger, `45730 = 2690 * 17`)
Discover_-_Elastic
Before Elastic:
Screenshot 2024-09-24 at 13 42 18

PR #812, opened 2024-09-27
Fix count (quick)

trzysiek:
I'll add tests in a next PR https://github.com/QuesmaOrg/quesma/pull/793 For now, need a quick fix merged ASAP.

Before:
Screenshot 2024-09-27 at 13 16 56
After:
Screenshot 2024-09-27 at 13 17 04

### DIFF
--- a/quesma/queryparser/query_translator.go
+++ b/quesma/queryparser/query_translator.go
@@ -12,7 +12,6 @@ import (
 	"quesma/quesma/config"
 	"quesma/schema"
 	"quesma/util"
-	"strings"
 )
 
 type JsonMap = map[string]interface{}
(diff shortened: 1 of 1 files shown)


quesma/queryparser/query_translator.go
```diff
@@ -178,44 +177,19 @@ func (cw *ClickhouseQueryTranslator) makeTotalCount(queries []*model.Query, resu
...
-							case uint64:
-								totalCount += int(v)
-							case int:
-								totalCount += v
-							case int64:
-								totalCount += int(v)
-							default:
-								logger.ErrorWithCtx(cw.Ctx).Msgf("Unknown type of count %v %t", v, v)
-							}
+				firstRow := results[queryIdx][0]
```
nablaone, line 180: We should check if the array has at least 1 element.

nablaone (approved):

quesma/queryparser/query_translator.go
trzysiek: we're doing it above :P

PR #815, opened 2024-09-30
No ingest setup

nablaone:
Changes:
- we can add  a single query pipeline now
- ingests attempts will end up with an error
- common table will not be created on the startup


quesma/quesma/config/config_v2.go
```diff
@@ -529,6 +582,10 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 
 			conf.IndexConfig[indexName] = processedConfig
 		}
+
+		conf.EnableIngest = true
+		conf.IngestStatistics = c.IngestStatistics
```
avelanarius, line 587:
```suggestion
		conf.IngestStatistics = true
```

avelanarius: (In my opinion it should be always turned on and not rely on user's configuration (at least yet))

quesma/quesma/config/config_v2.go
```diff
@@ -456,24 +454,79 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
...
+			// Handle default index configuration
+			defaultConfig := conf.IndexConfig[DefaultWildcardIndexName]
+			if len(defaultConfig.QueryTarget) != 1 {
+				errAcc = multierror.Append(errAcc, fmt.Errorf("the target configuration of default index ('%s') of query processor is not currently supported", DefaultWildcardIndexName))
+			}
+			if defaultConfig.QueryTarget[0] == ClickhouseTarget {
+				conf.AutodiscoveryEnabled = true
+			}
+
+			delete(conf.IndexConfig, DefaultWildcardIndexName)
```
mieciu, line 523:
```suggestion

```

mieciu (approved):

PR #816, opened 2024-09-30
Remove the need for SQL-type backend connector in transparent proxy mode

mieciu:
As in the description. The way I did it is quite ugly (esp. the `(l *LogManager) IsInTransparentProxyMode()`), but for now I think it's a decent compromise.

### DIFF
--- a/quesma/quesma/config/config_v2.go
+++ b/quesma/quesma/config/config_v2.go
@@ -469,7 +469,7 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 	conf.IngestStatistics = c.IngestStatistics
 	conf.AutodiscoveryEnabled = false
 	conf.Connectors = make(map[string]RelationalDbConfiguration)
-	relDBConn, connType, err := c.getRelationalDBConf()
+	relDBConn, connType, relationalDBErr := c.getRelationalDBConf()
 
 	isSinglePipeline, isDualPipeline := c.getPipelinesType()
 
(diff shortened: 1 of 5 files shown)


quesma/quesma/config/config_v2.go
```diff
@@ -573,6 +561,29 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
...
+	if err != nil && !conf.TransparentProxy {
+		errAcc = multierror.Append(errAcc, err)
+	} else if err != nil && conf.TransparentProxy {
+		relDBConn := RelationalDbConfiguration{
+			ConnectorType: ClickHouseOSBackendConnectorName,
+			Url: &Url{
+				Host: "localhost",
+			},
+		}
+		conf.Connectors["mock-for-transparent-proxy"] = relDBConn
```
avelanarius, line 574: Are there any spammy logs due to that fake DB configuration?

quesma/quesma/config/config_v2.go
```diff
@@ -573,6 +561,29 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 	}
 
 END:
+
+	if err != nil && !conf.TransparentProxy {
```
avelanarius, line 565: Nit: since the distance between setting `err` on line 472 (`relDBConn, connType, err := c.getRelationalDBConf()`) and this actual use of `err` is so large, maybe it'd be better to rename `err` to `relationalDBErr`.

mieciu: good call https://github.com/QuesmaOrg/quesma/pull/816/commits/020b45d1bd59ae6caabe51ffee56f1a463a0af80

avelanarius (approved):

quesma/quesma/config/config_v2.go
```diff
@@ -573,6 +561,29 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 	}
 
 END:
+
```
jakozaur, line 564: Personally, not a great fun of using labels `END` that way. This could be a function/method call instead.

jakozaur (approved):
One comment.

PR #819, opened 2024-09-30
Change default log level to `INFO`

avelanarius:
Before this change if the user did not configure the log level in the configuration it would default to `DEBUG` (as this is the first entry in `DebugLevel` enum). Debug log level is very verbose and it's not a great experience for a user (who tried a minimal config).

Therefore set log level to `INFO` if none was provided in the loaded config.


mieciu (approved):
👍

jakozaur (approved):

PR #822, opened 2024-10-01
Remove requirement of defining indexes in both processors, bring back required `*` configuration

avelanarius:
Remove the requirement that the user has to configure some index in both ingest and query processors. After this change, the user can now define an index only in ingest or only in query processor.

Under the hood, the translated V1 index configuration will have an ingest target set to `DefaultIngestTarget` if index is missing from ingest processor configuration or query target set to `DefaultQueryTarget` if index is missing from query processor configuration.

Bring back required `*` configuration. The only supported configuration is `"*": target: [elastic]`. This will make it easier for the users to understand the configuration, instead of relying on some "magic" fallback behavior.

Additionally improve the logging of index configurations (printed at the start of Quesma).

### DIFF
--- a/ci/quesma/config/ci-config.yaml
+++ b/ci/quesma/config/ci-config.yaml
@@ -37,6 +37,8 @@ processors:
                 type: text
         windows_logs:
           target: [ my-clickhouse-data-source ]
+        "*":
+          target: [ my-minimal-elasticsearch ]
   - name: p2
     type: quesma-v1-processor-ingest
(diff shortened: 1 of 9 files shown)


mieciu (approved):
:shipit:

avelanarius:
Adding WIP re-adding of `*`. Still testing that.

mieciu (approved):
:shipit::shipit:

PR #824, opened 2024-10-01
Introduce `nightly` Docker tag

mieciu:
Changes to how we build/publish Quesma Docker images:
* what used to be `quesma/quesma:latest` now becomes a pointer to **latest released version**
* introducing `quesma/quesma:nightly` image which becomes the most recent (not necessarily released/stable) version, always built and pushed from the `main` branch
* removing image pushes to GCR for  `quesma/quesma:nightly`. Release images will stay, as GCP's products don't support pulling from DockerHub (😢)

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   quesma:
     build: ../quesma
-    image: quesma:latest
+    image: quesma:nigthly
     environment:
       - QUESMA_CONFIG_FILE=/mnt/ci-config.yaml
       - QUESMA_logging_path=/var/quesma/logs
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   quesma:
     build: ../quesma
-    image: quesma:nigthly
+    image: quesma:nightly
     environment:
       - QUESMA_CONFIG_FILE=/mnt/ci-config.yaml
       - QUESMA_logging_path=/var/quesma/logs
--- a/ci/e2e.yml
+++ b/ci/e2e.yml
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   quesma:
     build: ../quesma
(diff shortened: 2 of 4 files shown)


ci/ci.yml
```diff
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   quesma:
     build: ../quesma
-    image: quesma:latest
+    image: quesma:nigthly
```
avelanarius, line 5: Not 100% sure about it, but I think this is incorrect. 

In our tests pipeline we build `quesma:latest` (here: https://github.com/QuesmaOrg/quesma/blob/260002c19ac40d032293051ea8fecdf72f311e11/.github/workflows/pipeline.yml#L124-L140) and then we start up `ci/ci.yml` with that image (https://github.com/QuesmaOrg/quesma/blob/260002c19ac40d032293051ea8fecdf72f311e11/.github/workflows/pipeline.yml#L168). Wouldn't that launch a released nightly image instead of just built `quesma:latest`?

avelanarius, line 5: typo: `nigthly` instead of `nightly`

ci/e2e.yml
```diff
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   quesma:
     build: ../quesma
-    image: quesma:latest
+    image: quesma:nigthly
```
avelanarius, line 5: typo: `nigthly` instead of `nightly`

ci/ci.yml
avelanarius: OK, now I see that `ci/ci.yml` always rebuilds it (because it uses `quesma:nightly` instead of `quesma/quesma:nightly`).

mieciu: Yea but it's all local - hopefully sometime in the future we'll become an official image (instead of `quesma/quesma`) and this becomes a concern 😉

avelanarius (approved):

jakozaur (approved):

PR #825, opened 2024-10-01
Fix double JSON marshalling in elasticsearch client's Request*()

avelanarius:
`Request()` and `RequestWithHeaders()` methods of elasticsearch client always "JSON marshalled" the incoming payload (body). However all users of those methods actually already gave a JSON marshalled payload! This caused the request body to be double marshalled, resulting in some strange base64 blob, which broke our double writes (sending bulk to Elastic).

Fix the problem by NOT marshalling in `Request*()` methods. Improve the `client_test.go` to validate that the client did not tamper with the body.

### DIFF
--- a/quesma/elasticsearch/client.go
+++ b/quesma/elasticsearch/client.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


avelanarius:
Thanks to @nablaone for finding this bug!

nablaone (approved):
LGTM

PR #826, opened 2024-10-01
Transparent proxy fixes

jakozaur:
My trial run found several DX issues and addressed them
- we inspect packages in transparent proxy and write weird errors when JSON or NJSON is here
  - disabled for now
- we show ClickHouse DB error
  - don't do try on moc error
  - in fact remove `mock-for-transparent-proxy` as this is confusing
- A/B logs are there even if there is no test
  - do detection on that
  
Along the way added constant and removed some dead code.


nablaone (approved):
LGTM

PR #827, opened 2024-10-01
Fix parsing the configuration

nablaone:


quesma/quesma/config/config_v2.go
```diff
@@ -630,6 +630,7 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 
 END:
 
+
 	if relationalDBErr != nil && !conf.TransparentProxy {
```
avelanarius, line 634:
```suggestion
	if relationalDBErr != nil && !conf.TransparentProxy {
```

avelanarius: for gofmt to not complain

PR #829, opened 2024-10-01
Fix completion_status bug

jakozaur:
We return `completion_status` HTTP code which causes bug in certain version of Kibana.


jakozaur:
Not enough though.

PR #832, opened 2024-10-02
Fix problems with dates part 1

trzysiek:
Query from the test caused Quesma to crash without any error message, as it'd run of out memory in 0,1s or so. You can see it below:
Screenshot 2024-10-02 at 15 33 54
Screenshot 2024-10-02 at 15 34 09
It was a small bug when adding empty rows. Usually keys in `date_histogram` are `x, x+1, x+2,...`, when we're using `timestamp / duration` as a key, so number of added empty rows was kind of limited. But in `calendar_interval` for big intervals, like 1 week, keys are timestamps in millisecond, so we were adding a new row for each millisecond, exhausting all the memory very quickly.
Fixed here by considering the second option.

### DIFF
--- a/quesma/kibana/intervals.go
+++ b/quesma/kibana/intervals.go
@@ -11,6 +11,21 @@ import (
 func ParseInterval(fixedInterval string) (time.Duration, error) {
 	var unit time.Duration
 
+	switch fixedInterval {
+	case "minute":
+		return time.Minute, nil
+	case "hour":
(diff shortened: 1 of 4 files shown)


jakozaur:
@trzysiek perhaps we should add safeguard to not add more than X missing timeslots (like 1000). It looks like we still may hit this issue.

jakozaur (approved):

PR #835, opened 2024-10-03
Fix Hydrolix dates 1

trzysiek:
This eliminates an error we encountered with Hydrolix.
```
Q3006: Unspecified database error.  clickhouse: query failed. err: code: 169, message: Key expression contains comparison between inconvertible types: DateTime64(3) and Float64 inside reqTimeSec >= 1727858503270
```
It's possible to compare dates and floats in Clickhouse from 2022 year, but in 2021 it wasn't possible, and Hydrolix must use some earlier version, so I created a solution which works in both scenarios

I move date parsing from Clickhouse to Quesma, removing usage of Clickhouse's `parseDateTimeBestEffort()`. Now we parse date ourselves, and use `toDateTime64()` to compare this date with date field (checked, it works fine also for fields of basic `DateTime` type, not `DateTime64`, even in Clickhouse 4 years ago). Elastic has like 50 different date formats available, so I doubt all of them are available in Clickhouse or other databases, so it's a move in the right direction, I guess.

After:
Screenshot 2024-10-06 at 17 15 22
Timestamps are fine: there's `"gte": 1727858503270` in the request, which is `Wed Oct 02 2024 08:41:43 GMT+0000`, so histogram starts fine from `8:41:30`:
Screenshot 2024-10-06 at 17 15 49


jakozaur (commented):
Great direction?

How did you find this and concluded it is Hydrolix related? I looked into log lines and just hit this error in dev env by me and Rafal.

quesma/queryparser/query_parser.go
```diff
@@ -789,84 +786,81 @@ func (cw *ClickhouseQueryTranslator) parseRange(queryMap QueryMap) model.SimpleQ
...
-			if !isDatetimeInDefaultFormat {
-				timeFormatFuncName = "toUnixTimestamp64Milli"
-				finalLHS = model.NewFunction(timeFormatFuncName, model.NewColumnRef(field))
-			} else {
-				switch fieldType {
-				case clickhouse.DateTime64, clickhouse.DateTime:
-					if dateTime, ok := v.(string); ok {
-						// if it's a date, we need to parse it to Clickhouse's DateTime format
-						// how to check if it does not contain date math expression?
-						if _, err := iso8601.ParseString(dateTime); err == nil {
```
jakozaur, line 822: If we are not using `iso8601`, we should drop it from dependencies.

jakozaur (commented):
We likely need more simple tests for this date logic.

trzysiek:
Hmm, I just knew that it always used to work fine in Clickhouse, and in Hydrolix it didn't, so I:
* googled it and found that in Clickhouse this functionality was added somewhere in 2021-2022
* found solution that works both now and 3 years ago, implemented it and checked it indeed works in Hydrolix too.

jakozaur (commented):
Please use int64 instead of float64. IDeally use:
fromUnixTimestamp64Milli

jakozaur (approved):
Applied my changes to use int64 instead of float64.

PR #836, opened 2024-10-03
Adding admin console auth

mieciu:
Adding admin console auth. This might seem really funny at glance with the whole empty `NewElasticsearchAuthProvider()` which is essentially used only to add two routes, but in the future adding any oAuth provider should be trivial.

In fact, I was able to create Google oAuth app and log in using my credentials.

For now however, we have a nice login screen asking for our Elasticsearch credentials. The whole logic is really in `HandleElasticsearchLogin`.

At this moment, **if Elasticsearch cluster allows unauthenticated access, Quesma disables admin console auth.**

image (20)


quesma/quesma/ui/console_routes.go
```diff
@@ -182,7 +235,28 @@ func (qmc *QuesmaManagementConsole) initPprof(router *mux.Router) {
...
+
+func authMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		//if !authEnabled {
+		//	next.ServeHTTP(w, r)
+		//	return
+		//}
+		session, err := store.Get(r, quesmaSessionName)
+		userID, ok := session.Values["userID"].(string)
+		if !ok || userID == "" || err != nil {
```
nablaone, line 248: It will be tricky to guess "why it redirects me to the login form". We should log and error here.

quesma/quesma/ui/login.go
```diff
@@ -0,0 +1,60 @@
...
+	buffer.Html(`</form>`)
+	buffer.Html(`</div>`)
+	buffer.Html(`</div>`)
+	buffer.Html(`</body>`)
+	buffer.Html(`</html>`)
+	return buffer.Bytes()
+}
+
+func (qmc *QuesmaManagementConsole) HandleElasticsearchLogin(writer http.ResponseWriter, req *http.Request) {
+	if req.Method == http.MethodGet {
```
nablaone, line 37: We should check if the user is logged in already here.

quesma/quesma/ui/login.go
```diff
@@ -0,0 +1,60 @@
...
+		writer.Header().Set("Content-Type", "text/html")
+		writer.Write(qmc.generateLoginForm())
+	} else if req.Method == http.MethodPost {
+		username := req.FormValue("username")
+		password := req.FormValue("password")
+		if qmc.isValidElasticsearchUser(username, password) {
+			session, _ := store.Get(req, quesmaSessionName)
```
nablaone, line 44: Add an error log here. It's tricky to debug login issues.

jakozaur:
> At this moment, if Elasticsearch cluster allows unauthenticated access, Quesma disables admin console auth.
Not sure, currently unauthenticated access is useful for dev or evaluation mode, maybe we should keep it that way.

quesma/quesma/ui/login.go
mieciu: https://github.com/QuesmaOrg/quesma/pull/836/commits/2dee16ae8063e7e2d755e8925b49bd4752a6e2ae

quesma/quesma/ui/login.go
mieciu: https://github.com/QuesmaOrg/quesma/pull/836/commits/9bff908d6cddb3483472e0e4cdbdf3cd44f81c51

quesma/quesma/ui/console_routes.go
mieciu: I'd expect this to be pretty usual behavior, but agree that `logger.Warn().Msgf("User not authenticated, redirecting to login page")` is worth having there

nablaone (approved):

PR #837, opened 2024-10-03
Quickfix build-image

trzysiek:
@mieciu said it should be changed


bin/build-image.sh
```diff
@@ -2,4 +2,4 @@
 # Rebuilds the Quesma Docker image and tags it as quesma:latest
```
avelanarius, line 2:
```suggestion
# Rebuilds the Quesma Docker image and tags it as quesma:nightly
```

avelanarius (approved):

jakozaur (approved):

PR #841, opened 2024-10-03
Fix histogram when `interval==1`

trzysiek:
When `interval` is any other than `1`, we receive a float, so no change needed (as we do `(field / interval) * interval` and it's a float even if `interval` is an integer, like `2` (tested))
When `interval` is `1`, we get any type the field is in DB, as we just do `SELECT field`, so we need to check all int/float types. Added that here.
Before:
Screenshot 2024-10-03 at 14 27 30
Screenshot 2024-10-03 at 14 27 23
After:
Screenshot 2024-10-03 at 14 50 38

### DIFF
--- a/quesma/model/bucket_aggregations/histogram.go
+++ b/quesma/model/bucket_aggregations/histogram.go
@@ -112,8 +112,8 @@ func (query *HistogramRowsTransformer) getKeyFloat64(row model.QueryResultRow) (
 // we don't know the type
 func (query *HistogramRowsTransformer) getKeyUnknownType(row model.QueryResultRow) (float64, bool) {
 	val := row.Cols[len(row.Cols)-2].Value
-	if valTyped, ok := util.ExtractNumeric64Maybe(val); ok {
-		return valTyped, true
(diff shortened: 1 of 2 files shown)


quesma/model/bucket_aggregations/histogram.go
```diff
@@ -74,6 +74,12 @@ func (query *HistogramRowsTransformer) Transform(ctx context.Context, rowsFromDB
 	}
 	postprocessedRows := make([]model.QueryResultRow, 0, len(rowsFromDB))
 	postprocessedRows = append(postprocessedRows, rowsFromDB[0])
+
+	getKey := query.getKeyFloat64
+	if query.interval == 1.0 {
+		getKey = query.getKeyUnknownType
+	}
```
jakozaur, line 81: Shouldn't be that:
```
var getKey float64
if query.interval == 1.0 {
  getKey = query.getKeyUnknownType
} else {
  getKey := query.getKeyFloat64
}
```

jakozaur (approved):
Not sure if I get it.

PR #846, opened 2024-10-04
Integration test "framework"

mieciu:
How to run this:
```
cd ci/it
go test -v
```

What I really wanted to achieve here is to provide some sort of separation between test infra setup and the tests themselves.

The ITs are listed in `ci/it/integration_test.go`.
Each test is a separate struct living a a separate file. The test struct exposes handy methods like `RequestToQuesma()`, `RequestToElasticsearch()`, `ExecuteClickHouseStatement()` and `a.ExecuteClickHouseQuery()`.
The whole test logic lives in overriden `RunTests` function.  
Under the hood, there's whole machinery which spins up required docker containers and creates dedicated Quesma config file based on the provided template. Good news is, one does not have to touch it in order to run and contribute to the tests 😉


ci/it/configs/quesma-as-transparent-proxy.yml.template
```diff
@@ -0,0 +1,31 @@
+installationID: "PAN Z WAMI I Z DU"
```
avelanarius, line 1: "I Z DU" could be misinterpreted by some people :(

ci/it/testcases/test_reading_clickhouse_tables.go
```diff
@@ -0,0 +1,59 @@
...
+
+	insertRowsQuery := "INSERT INTO test_table (id, name) VALUES (1, 'Alice'), (2, 'Bob'), (3, 'Charlie')"
+	if _, err := a.ExecuteClickHouseStatement(ctx, insertRowsQuery); err != nil {
+		return err
+	}
+
+	// This returns 500 Internal Server Error, but will be tackled in separate PR.
+	// (The table has not yet been discovered by Quesma )
+	// ERR quesma/quesma/quesma.go:198 > quesma request failed: Q2002: Missing table. Table: test_table: can't load test_table table opaque_id= path=/test_table/_search reason="Missing table." request_id=01926654-b214-7e1d-944a-a7545cd7d419
+	resp, err = a.RequestToQuesma(ctx, "GET", "/test_table/_search", []byte(`{"query": {"match_all": {}}}`))
```
avelanarius, line 51: An idea for the future: we could also test if Elastic's go client can connect to Quesma and query it correctly

avelanarius (approved):

PR #847, opened 2024-10-05
Fix dates in pipeline aggregations

trzysiek:
Unskips one test.

We had a problem when generating results for pipeline aggregations, because if bucket's key was a date, pipeline aggregation only had what it received from DB (usually some random int, result of `date_field.ToUnixTimestamp() / interval_duration`, let's call it `X` (`originalKey` in the new code)), and had no idea how to transform it into a valid date in response.

I fixed that by connecting pipeline aggregation to its parent bucket aggregation, so now pipeline aggregation knows that its parent is a date_histogram, and if it is, it asks it to transform `X -> proper date`


jakozaur (approved):
More tests would be nice.

PR #848, opened 2024-10-05
small improvement `top_hits`

trzysiek:
* Fixes some small problem with not having count in `top_hits` pancake. We can't get it (at least easily) in our logic, as you can't do `SELECT field, count()`. But it doesn't seem to be a problem, it's of course present in some other pancake and it's only taken from there => seems working fine.
* Add some better support for what fields we want to return in `top_hits`. Before it had to be an array `params[_source][includes]`, now it can also. be a string `params[_source]`. It's not 100% support for this `_source` field, but a bit better.

### DIFF
--- a/quesma/model/bucket_aggregations/range.go
+++ b/quesma/model/bucket_aggregations/range.go
@@ -138,7 +138,7 @@ func (query Range) String() string {
 func (query Range) responseForInterval(interval Interval, value any) model.JsonMap {
 	response := model.JsonMap{}
 	if value != nil {
-		// sometimes we may not have count and it's ok (it's only in some other pancake)
+		// occasionally we may not have count (e.g. top_hits) and it's ok
 		response["doc_count"] = value
 	}
 	if !interval.IsOpeningBoundInfinite() {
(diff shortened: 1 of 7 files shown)


quesma/queryparser/pancake_json_rendering.go
```diff
@@ -27,6 +27,7 @@ func newPancakeJSONRenderer(ctx context.Context) *pancakeJSONRenderer {
 }
 
 func (p *pancakeJSONRenderer) selectMetricRows(metricName string, rows []model.QueryResultRow) (result []model.QueryResultRow) {
+	fmt.Println("selectt", metricName, rows)
```
jakozaur, line 30: This line should be removed.

quesma/model/bucket_aggregations/range.go
```diff
@@ -170,7 +172,11 @@ func (query Range) CombinatorGroups() (result []CombinatorGroup) {
 
 func (query Range) CombinatorTranslateSqlResponseToJson(subGroup CombinatorGroup, rows []model.QueryResultRow) model.JsonMap {
 	interval := query.Intervals[subGroup.idx]
-	count := rows[0].Cols[len(rows[0].Cols)-1].Value
+	var count any
+	if len(rows[0].Cols) > 0 {
+		// sometimes we may not have count and it's ok (it's only in some other pancake)
```
jakozaur, line 177: We may not have counted in `top_hits`, but most of the time, we should have it.

jakozaur (commented):
Looks good, let me removed one `println` and improve one fix.

jakozaur (approved):

PR #850, opened 2024-10-06
Add empty rows even if no `min_doc_count=0`

trzysiek:
Unskips a recently added test.

Basically, for `(date_)histogram`, we normally run our `RowsTransformer` only when `min_doc_count=0`. But for some pipeline aggregations, like `cumulative sum` or `derivative` we need to run it anyway for dashboard to look pretty.

There's general division of pipeline aggregations into parent/sibling (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline.html) Didn't use that before, but those parent ones seem to be e.g. the ones where we need to do that transformation, so introduced this division here.

Improved a few tests, added some empty rows in some to test this behaviour + fixed a few slightly incorrect (because of this behaviour) ones

Before:
Screenshot 2024-10-06 at 13 09 42
After: (same as Elastic)
Screenshot 2024-10-06 at 13 09 59


quesma/testdata/opensearch-visualize/pipeline_aggregation_requests.go
```diff
@@ -1766,17 +1913,9 @@ var PipelineAggregationTests = []testdata.AggregationTestCase{
...
 				model.NewQueryResultCol("aggr__2__count", uint64(2)),
 			}},
-			{Cols: []model.QueryResultCol{
-				model.NewQueryResultCol("aggr__2__key_0", int64(1714869600000/600000)),
-				model.NewQueryResultCol("aggr__2__count", uint64(0)),
-			}},
-			{Cols: []model.QueryResultCol{
-				model.NewQueryResultCol("aggr__2__key_0", int64(1714878600000/600000)),
-				model.NewQueryResultCol("aggr__2__count", uint64(0)),
-			}},
```
trzysiek, line 1779: was incorrect, DB wouldn't return rows with count 0

quesma/queryparser/pancake_transformer.go
```diff
@@ -267,13 +267,14 @@ func (a *pancakeTransformer) checkIfSupported(layers []*pancakeModelLayer) error
...
+			parentBucketLayer, layerIdx, err := a.findParentBucketLayer(layers[i:], pipeline.queryType)
 			if err != nil {
 				logger.WarnWithCtx(a.ctx).Err(err).Msg("could not find parent bucket layer")
 				continue
 			}
-			if parentBucketLayer.nextBucketAggregation != nil {
-				pipeline.queryType.SetParentBucketAggregation(parentBucketLayer.nextBucketAggregation.queryType)
+
+			if i+layerIdx > 0 {
+				pipeline.queryType.SetParentBucketAggregation(layers[i+layerIdx-1].nextBucketAggregation.queryType)
```
trzysiek, line 277: it was incorrect before, was usually giving one bucket aggregation too far

jakozaur (commented):
It's step in good direction, but not sure if it's correct. 

> You can control the order of the returned buckets using the order settings and filter the returned buckets based on a min_doc_count setting (by default all buckets between the first bucket that matches documents and the last one are returned).

I believe the logic might be slightly different.

jakozaur:
It might be correct, let's talk about it.

jakozaur (approved):

PR #853, opened 2024-10-07
Use ahead computed field encodings

pdelewski:
This PR is about 
1) Removing one of the shortcuts (i.e., removing the TODO) and reading from the schemaRegistry instead of directly using the FieldToColumnEncoder call. 
2) Additionally, removing buildCreateTableQueryNoOurFields, which was only used in tests.


quesma/schema/static_registry.go
```diff
@@ -39,3 +39,10 @@ func (e StaticRegistry) UpdateFieldEncodings(encodings map[FieldEncodingKey]Enco
 		e.FieldEncodings[k] = EncodedFieldName(v)
 	}
 }
+
```
nablaone, line 42: We should change all methods to be pointer receivers. These value receivers are tricky. Code looks "ok", but it don't work.

nablaone (approved):

quesma/schema/static_registry.go
pdelewski: You are right

PR #855, opened 2024-10-07
Adding GitHub Workflow for integration tests

mieciu:
Adding few tweaks so that test **run on CI** (wasn't that straightforward TBH) and are not flaky. 

Overall, those two ITs on GitHub worker have never took more than 3 minutes total, which I think as for IT is pretty decent.
image

For now, running only `quesma/quesma:latest` docker image upon request. Later we can discuss when/where we want them to be kicked off.

### DIFF
--- a/ci/it/testcases/utils.go
+++ b/ci/it/testcases/utils.go
@@ -18,7 +18,8 @@ const configTemplatesDir = "configs"
 
 func GetInternalDockerHost() string {
 	if check := os.Getenv("EXECUTING_ON_GITHUB_CI"); check != "" {
-		return "localhost"
+		//return "localhost"
+		return "host.testcontainers.internal"
 	}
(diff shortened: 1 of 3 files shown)


ci/it/testcases/utils.go
```diff
@@ -14,10 +15,15 @@ import (
 	"time"
 )
 
-// InternalDockerHost (`host.testcontainers.internal`) doesn't work for Docker Desktop for Mac.
-const InternalDockerHost = "host.docker.internal"
 const configTemplatesDir = "configs"
 
+func GetInternalDockerHost() string {
+	if check := os.Getenv("EXECUTING_ON_GITHUB_CI"); check != "" {
+		return "localhost-for-github-ci"
```
pdelewski, line 22: It would be good to extract these strings as constants

pdelewski (approved):

PR #862, opened 2024-10-09
Extract `build-quesma-docker-image` to a reusable GitHub Actions workflow

avelanarius:
While working on an upcoming PR, I noticed that `build-quesma-docker-image` step in GitHub Actions workflow was duplicated in 3 places... and my PR would add a 4th copy!

Therefore extract the `build-quesma-docker-image` step into a "reusable workflow" (`workflow_call` trigger) which can be called from other workflows, avoiding duplication. All the behaviors of this step stay the same - for now the "build+push" step and "build+export" step are separate to make sure the behavior stays the same.

Additionally `docker-image-latest.yml` is renamed to `nightly-docker-build-and-push.yml` in order to avoid confusion with the new `build-quesma-docker-image.yml` file.


avelanarius:
Tested it on my fork: https://github.com/avelanarius/quesma/commits/main/

.github/workflows/build-quesma-docker-image.yml
```diff
@@ -0,0 +1,88 @@
...
+            quesma/quesma:nightly
+          push: false
+          outputs: type=docker,dest=/tmp/image.tar
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+        env:
+          DOCKER_BUILD_SUMMARY: false
+
+      # If we push the image, we build it for both amd64 and arm64 and don't export it
+      - name: Build and push
```
avelanarius, line 65: The reason there's this duplication of "Build and export" and "Build and push" steps is that the first one doesn't build for arm64 in order to save time (arm64 build takes a lot of time on GitHub Actions, it's probably running emulated ARM on x86). But on the flip side, this preserves the exactly same behavior as before.

mieciu: I think this GitHub comment should land in the yaml 😉

mieciu (approved):
nice! 💪

.github/workflows/build-quesma-docker-image.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: docker/login-action@v3
+        if: ${{ inputs.PUSH }}
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PAT }}
+
+      - name: Set the build date
+        run: echo QUESMA_BUILD_DATE=$(git --no-pager log -1 --date=format:'%Y-%m-%d' --format="%ad")  >> $GITHUB_ENV
+
+      # If we don't push the image, we build it only for the host architecture and export it to .tar file
```
avelanarius, line 48:
```suggestion
      # If we don't push the image, we build it only for the host architecture and export it to .tar file
      # This step is similar to "Build and push", but avoids building arm64 image, which takes
      # a lot of time on GitHub Actions.
```

PR #865, opened 2024-10-09
Run integration tests against `nightly` (not `latest`)

avelanarius:
The `latest` tag of `quesma/quesma` is the latest released stable release of Quesma, while `nightly` is the newest build of Quesma. Integration tests used to run against `quesma/quesma:latest` and not the last `nightly` build of Quesma, which meant that we wouldn't catch some failures before release.

Fix this issue by running the integration tests against `nightly` and always building the newest `quesma/quesma:nightly` in the `integration-test.yml` GitHub workflow.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -10,9 +10,14 @@ on:
         description: 'Commit hash to run the tests'
         required: true
 jobs:
+  build-quesma-docker-image:
+    uses: ./.github/workflows/build-quesma-docker-image.yml
+    with:
+      REF: ${{inputs.GIT_REF}}
 
   integration-test-run:
     runs-on: ubuntu-latest
+    needs: [build-quesma-docker-image]
     steps:
       - uses: actions/checkout@v4
         with:     ## @TODO REMOVE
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -10,6 +10,42 @@ on:
...
+          cache-to: type=gha,mode=max
+        env:
+          DOCKER_BUILD_SUMMARY: false
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.module }}
+          path: /tmp/image.tar
+          retention-days: 1
```
avelanarius, line 48: Blatant copy-paste from another workflow. This copy-paste will get fixed by: https://github.com/QuesmaOrg/quesma/pull/862

mieciu: maybe let's merge #862 and apply it then ?

avelanarius: 👍

avelanarius:
In the amended commit I removed the duplication of Docker building step and fixed a couple small bugs.

Tested the integration test workflow here: https://github.com/avelanarius/quesma/actions/workflows/integration-tests.yml
(run 5 is a run on latest master, run 4 is a run on a branch with deliberate panic and the test caught that - used correct new image, ignore runs 1-3)

.github/workflows/integration-tests.yml
avelanarius: Fixed in the latest version

mieciu (approved):

PR #868, opened 2024-10-10
Get secrets (Docker login credentials) in `build-quesma-docker-image.yml`

avelanarius:
A hot fix after 0c220f988a947ca63f3ca51c65a4ba05a2162f13. After it was merged, the reusable workflow ("child workflow") doesn't get the secrets (docker credentials) by default, resulting in a job failure.

Fix the issue by specifying the needed secrets for the reusable workflow.

The secrets are not required if `PUSH` doesn't happen and only the `nightly-docker-build-and-push.yml` workflow actually passes those secrets.

### DIFF
--- a/.github/workflows/build-quesma-docker-image.yml
+++ b/.github/workflows/build-quesma-docker-image.yml
@@ -16,6 +16,11 @@ on:
         default: false
         required: false
         type: boolean
+    secrets:
+      DOCKER_USER:
+        required: false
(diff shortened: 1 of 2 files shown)


avelanarius:
Before the fix the job failed with this error:

image

Now, the job gets the credentials (I put some dummy incorrect ones on my fork for testing):
image

mieciu (approved):

PR #870, opened 2024-10-10
Table resolver - apply decision

nablaone:
This is second part of the table resolver introduction. 

In this PR:
1. Decisions made by resolved are applied. 
2. Some rules were simplified
3. Add test.
4. Add endpoint `/index:/_quesma_table_resolver` to get a decision (it will be used in e2e tests)
```
 curl 'http://localhost:8080/logs*/_quesma_table_resolver
````

Limitations:
- Table name overrides are handled as before. 
- Rules may be incorrect. Tests are passing. But this is complex rewrite.

### CODE REVIEW

File: quesma/table_resolver/table_resolver.go [hunk 1]
```diff
@@ -278,10 +279,10 @@ func NewTableResolver(quesmaConf config.QuesmaConfiguration, discovery clickhous
 			decisionLadder: []basicResolver{
```
pdelewski, line 279: Having this decisionLadder feels a bit overengineered to me. Why not just merge all these functions into one. This would improve readability IMO as we could read it from top to bottom

pdelewski: We can fix it later. I just think we don’t need this level of generality.

quesma/quesma/matchers.go
```diff
@@ -36,100 +34,58 @@ func matchedAgainstBulkBody(configuration *config.QuesmaConfiguration, tableReso
...
-
-		indexConfig, exists := cfg.IndexConfig[req.Params["index"]]
-		if queryPath {
-			return exists && indexConfig.IsClickhouseQueryEnabled()
-		} else {
-			return exists && (indexConfig.IsClickhouseIngestEnabled() || indexConfig.IsIngestDisabled())
+		for _, connector := range decision.UseConnectors {
+			if _, ok := connector.(*table_resolver.ConnectorDecisionClickhouse); ok {
+				return mux.MatchResult{Matched: true, Decision: decision}
+			}
```
avelanarius, line 71: Similarly to the previous code, this function should return `Matched: true` if ingest to an index is closed - that's the only way Quesma has chance to return an error to the client - now I think it'll just forward it to Elastic instead of blocking.

nablaone: Done

nablaone:
> I think we should merge it, but the integration tests are failing after this PR: https://github.com/QuesmaOrg/quesma/actions/runs/11362688004

It's ok now.
https://github.com/QuesmaOrg/quesma/actions/runs/11365696747

avelanarius (approved):

PR #871, opened 2024-10-10
Fix `nil` pointer panic in `table_discovery.go`

avelanarius:
After 7611ebac683e5e2fc326ea3afac8f80b3abf37e1, the table resolver always asks table discovery to refresh table information. However, in the transparent proxy case, Quesma doesn't make any connections to ClickHouse, therefore `td.dbConnPool` can be `nil`, resulting in `nil` pointer panic.

Add missing `nil` check, the existing code now can gracefully handle this situation. 

In the future we might improve Quesma's main to not construct `table_discovery` if there is no ClickHouse connector.


avelanarius:
@mieciu's integration tests caught this error: https://github.com/QuesmaOrg/quesma/actions/runs/11272817964/job/31348554612

PR #873, opened 2024-10-10
Allow env variables

jakozaur:
While creating demo env variable, I noticed it is impossible to pass credentials by variable. Unfortunately, `koan` does not support overriding YAML arrays directly, so I had to add this functionality.

### DIFF
--- a/quesma/quesma/config/config_v2_test.go
+++ b/quesma/quesma/config/config_v2_test.go
@@ -16,8 +16,15 @@ func TestQuesmaConfigurationLoading(t *testing.T) {
 
 	logLevelPassedAsEnvVar := "debug"
 	licenseKeyPassedAsEnvVar := "arbitraty-license-key"
-	os.Setenv("QUESMA_logging_level", logLevelPassedAsEnvVar) // overrides what's in the config file
-	os.Setenv("QUESMA_licenseKey", licenseKeyPassedAsEnvVar)  // overrides what's in the config file
+	os.Setenv("QUESMA_logging_level", logLevelPassedAsEnvVar)   // overrides what's in the config file
+	os.Setenv("QUESMA_licenseKey", licenseKeyPassedAsEnvVar)    // overrides what's in the config file
+	os.Setenv("QUESMA_backendConnectors_1_config_user", "user") // overrides what's in the config file
+	t.Cleanup(func() {
+		os.Unsetenv(configFileLocationEnvVar)
+		os.Unsetenv("QUESMA_logging_level")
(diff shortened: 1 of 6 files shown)


quesma/quesma/config/config_v2_test.go
```diff
@@ -16,8 +16,15 @@ func TestQuesmaConfigurationLoading(t *testing.T) {
 
 	logLevelPassedAsEnvVar := "debug"
 	licenseKeyPassedAsEnvVar := "arbitraty-license-key"
-	os.Setenv("QUESMA_logging_level", logLevelPassedAsEnvVar) // overrides what's in the config file
-	os.Setenv("QUESMA_licenseKey", licenseKeyPassedAsEnvVar)  // overrides what's in the config file
+	os.Setenv("QUESMA_logging_level", logLevelPassedAsEnvVar)   // overrides what's in the config file
+	os.Setenv("QUESMA_licenseKey", licenseKeyPassedAsEnvVar)    // overrides what's in the config file
+	os.Setenv("QUESMA_backendConnectors_1_config_user", "user") // overrides what's in the config file
```
avelanarius, line 21: It'd be cool if in addition to addressing backend connectors by index (`1`) you could address them by name, e.g. `QUESMA_backendConnectors_my-minimal-elasticsearch_config_user` (name from `name` field of connector)

But this probably will be messy if the name contains `_` separator.

mieciu: From that I'd expect:
```yaml
backendConnectors:
  my-minimal-elasticsearch:
    config:
      user: 
```
This goes against the syntax and introduces unnecessary ambiguity IMHO.

jakozaur: Ok, sounds a valid use case, though don't know how to implement that easily.

mieciu: So while imperfect, I think it's easier to understand the final result 😉

avelanarius: OK, I see. If there was an `backendConnectors` array already, we could deduce that `my-minimal-elasticsearch` is an "index" into that array, but if the configuration doesn't have `backendConnectors` at all we can't deduce that `backendConnectors` is an array.

I suggested this other approach since index into an array will get invalidated if the user adds another backend connector into the config - using a name would be more stable.

jakozaur: @mieciu @avelanarius I added support by name and this looks much better.

avelanarius (approved):

PR #878, opened 2024-10-11
Fixing quesma common table dynamic mapping case

pdelewski:
This PR fixes the issue where quesma_common_table is used, and the mapping contains additional fields that were not present during ingestion.

```
processors:
  - name: my-query-processor
    type: quesma-v1-processor-query
    config:
      indexes:
        kibana_sample_data_ecommerce:
          useCommonTable: true
          schemaOverrides:
            fields:
              "geoip.location":
                type: geo_point
```

Before :
image

After :
image

Additionally, it introduces:
- The AlterDDL struct to represent ALTER statements in object form instead of as a string. My plan is to do more refactoring around this in subsequent PR(s).
- A FieldOrigin property, which provides information about the source of the field's metadata.

### DIFF
--- a/quesma/ingest/processor.go
+++ b/quesma/ingest/processor.go
@@ -286,6 +286,13 @@ func getAttributesByArrayName(arrayName string,
 	return attributes
 }
 
+type AlterDDL struct {
+	tableName  string
+	columnName string
+	columnType string
+	comment    string
+}
+
 // This function generates ALTER TABLE commands for adding new columns
 // to the table based on the attributesMap and the table name
 // AttributesMap contains the attributes that are not part of the schema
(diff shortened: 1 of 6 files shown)


quesma/ingest/processor.go
```diff
@@ -286,6 +286,13 @@ func getAttributesByArrayName(arrayName string,
 	return attributes
 }
 
+type AlterDDL struct {
+	tableName  string
```
nablaone, line 290: +1

nablaone:
What will happen?
1.  Sth puts a mappings  (with field `foo`)
2. Quesma restarts
3. Sth starts ingesting (no `foo` field)

pdelewski:
I will try other idea to fix it

pdelewski:
Closing as duplicate

PR #885, opened 2024-10-14
Fixing quesma common table dynamic mapping case, second approach

pdelewski:
This is a second approach (1 - https://github.com/QuesmaOrg/quesma/pull/878) of fixing a problem where mappings contains more fields than ingest and sql queries are based on this knowledge. In this approach we are not altering table to have all fields that were added by dynamic mapping but we based on field origin information and building queries according to that.

image

### DIFF
--- a/quesma/quesma/schema_transformer.go
+++ b/quesma/quesma/schema_transformer.go
@@ -463,7 +463,9 @@ func (s *SchemaCheckPass) applyWildcardExpansion(indexSchema schema.Schema, quer
 
 		cols := make([]string, 0, len(indexSchema.Fields))
 		for _, col := range indexSchema.Fields {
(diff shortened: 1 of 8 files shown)


quesma/quesma/schema_transformer.go
```diff
@@ -491,7 +494,10 @@ func (s *SchemaCheckPass) applyFullTextField(indexSchema schema.Schema, query *m
 
 	for _, field := range indexSchema.Fields {
 		if field.Type.IsFullText() {
-			fullTextFields = append(fullTextFields, field.InternalPropertyName.AsString())
+			// Take only fields that are ingested
```
nablaone, line 497: OK. Got it.

nablaone (approved):

PR #895, opened 2024-10-15
Persistent storage for asyncs 3

trzysiek:
Probably last, shouldn't take much longer than 0,5 day to finish, I think.


quesma/persistence/elastic.go
```diff
@@ -118,11 +118,10 @@ func (p *ElasticJSONDatabase) List() ([]string, error) {
 	}`
 
 	resp, err := p.httpClient.Request(context.Background(), "GET", elasticsearchURL, []byte(query))
-
+	defer resp.Body.Close()
```
elszon, line 121: According to [doc](https://pkg.go.dev/net/http#Client.Do) if `err != nil` body might be `nil`. So it looks like nil dereferencing or do I miss something?

PR #901, opened 2024-10-22
Make secure cookies working

nablaone:
We must provide keys for session storage (gothic and guerillla) to work with SSL.  Both frameworks switch to secure cookies on SSL. 

The solution is to generate keys on on start. This is for now.  This is not a final solution.

### DIFF
--- a/quesma/quesma/ui/console_routes.go
+++ b/quesma/quesma/ui/console_routes.go
@@ -31,6 +31,10 @@ var uiFs embed.FS
 const quesmaSessionName = "quesma-session"
 
 func init() {
+
+	// Here we generate a random key for the session store
+	// TODO We should use a secure key from the environment on production.
+	// 32 - is a default key length, taken for example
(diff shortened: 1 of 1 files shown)


quesma/quesma/ui/console_routes.go
```diff
@@ -247,7 +252,9 @@ func (qmc *QuesmaManagementConsole) initPprof(router *mux.Router) {
 	router.HandleFunc("/debug/pprof/trace", pprof.Trace)
 }
 
-var store = sessions.NewCookieStore([]byte("test"))
+var authKey = securecookie.GenerateRandomKey(64)
```
pdelewski, line 255: Do these values `64`, `32` ​​have a special meaning? I would suggest to add comment

pdelewski (approved):

quesma/quesma/ui/console_routes.go
nablaone: Added

PR #905, opened 2024-10-23
Allow more configurations of targets for "*"

avelanarius:
Previously, a user was allowed to only put ElasticSearch connector as a target for `*` index configuration (describing the default/fallback behavior). This PR lifts that requirement and allows for any connectors to be put into ingest's `*` and at most any 1 connector into query's `*`.

`table_resolver` code is modified to now read the `*` configuration and use it for decision (instead of always falling back to Elastic).

This PR also adds back enabling autodiscovery via configuration - if `ClickHouse` target is listed in `*` targets, autodiscovery will be turned on.


quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -261,74 +261,71 @@ func sendToClickhouse(ctx context.Context, clickhouseDocumentsToInsert map[strin
 	for indexName, documents := range clickhouseDocumentsToInsert {
 		phoneHomeAgent.IngestCounters().Add(indexName, int64(len(documents)))
 
-		config.RunConfiguredIngest(ctx, cfg, indexName, make(types.JSON), func() error {
```
avelanarius, line 264: The code no longer uses `RunConfiguredIngest` since that function had it's own logic of checking if an index is enabled for ingest based on the configuration, but now it's handled by table_resolver instead earlier.

avelanarius:
Things TODO, but I think they can be addressed in later PRs:
- Integration tests of all scenarios
- `field_caps`/`_resolve` might not work correctly

quesma/quesma/config/config_v2.go
```diff
@@ -603,11 +609,38 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 		}
 
 		conf.IndexConfig = make(map[string]IndexConfiguration)
+
+		// Handle default index configuration
+		defaultConfig := queryProcessor.Config.IndexConfig[DefaultWildcardIndexName]
+		for _, target := range defaultConfig.Target {
```
pdelewski, line 615: Just for consistency, maybe better would be to change this for loop to be same as below one with the exception of using `queryProcessor` or change name from `defaultConfig`  to `queryDefaultConfig` to be more meaningful.

Potentially, we could also try to extract a common function

pdelewski:
Overall, LGTM :)

quesma/quesma/config/config_v2.go
avelanarius: > change name from defaultConfig to queryDefaultConfig

That's not possible - both loops share the same `defaultConfig`, one fills out `QueryTarget` of it, while the second one fills out `IngestTarget` of it.

> Potentially, we could also try to extract a common function

Agreed, I'll do it in the follow up PR once we have integration tests covering this case.

PR #911, opened 2024-10-23
A/B testing - use the ingest processor instead of calling loopback

nablaone:
Let's introduce `intest.Ingester` interface.  
We can use our ingest processor internally now.


pdelewski:
LGTM :)

pdelewski (approved):

PR #915, opened 2024-10-24
Use common table for '*' wildcard

pdelewski:
This PR is about `usingCommonTable` configuration property for `*` wildcard:

```
'*':
  useCommonTable: true
```

### DIFF
--- a/quesma/quesma/config/config.go
+++ b/quesma/quesma/config/config.go
@@ -44,12 +44,11 @@ type QuesmaConfiguration struct {
 	DisableAuth                bool
 	AutodiscoveryEnabled       bool
 
-	EnableIngest                 bool // this is computed from the configuration 2.0
-	CreateCommonTable            bool
-	CreateCommonTableForWildcard bool // the meaning of this is to create a common table for wildcard indexes
-	// and is a duplication for now due to some corner cases
-	DefaultIngestTarget []string
-	DefaultQueryTarget  []string
+	EnableIngest              bool // this is computed from the configuration 2.0
+	CreateCommonTable         bool
+	UseCommonTableForWildcard bool //the meaning of this is to use a common table for wildcard (default) indexes
+	DefaultIngestTarget       []string
+	DefaultQueryTarget        []string
 }
 
 func (c *QuesmaConfiguration) AliasFields(indexName string) map[string]string {
--- a/quesma/quesma/config/config.go
+++ b/quesma/quesma/config/config.go
@@ -44,9 +44,10 @@ type QuesmaConfiguration struct {
 	DisableAuth                bool
 	AutodiscoveryEnabled       bool
 
-	EnableIngest      bool // this is computed from the configuration 2.0
-	CreateCommonTable bool
-
+	EnableIngest                 bool // this is computed from the configuration 2.0
+	CreateCommonTable            bool
+	CreateCommonTableForWildcard bool // the meaning of this is to create a common table for wildcard indexes
+	// and is a duplication for now due to some corner cases
 	DefaultIngestTarget []string
 	DefaultQueryTarget  []string
 }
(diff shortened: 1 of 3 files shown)


quesma/table_resolver/rules.go
```diff
@@ -79,6 +79,7 @@ func makeDefaultWildcard(quesmaConf config.QuesmaConfiguration, pipeline string)
 			case config.ClickhouseTarget:
 				useConnectors = append(useConnectors, &ConnectorDecisionClickhouse{
 					ClickhouseTableName: input.source,
+					IsCommonTable:       quesmaConf.CreateCommonTable,
```
avelanarius, line 82:
```suggestion
					IsCommonTable:       quesmaConf.CreateCommonTableForWildcard,
```

quesma/quesma/config/config.go
```diff
@@ -44,9 +44,10 @@ type QuesmaConfiguration struct {
 	DisableAuth                bool
 	AutodiscoveryEnabled       bool
 
-	EnableIngest      bool // this is computed from the configuration 2.0
-	CreateCommonTable bool
-
+	EnableIngest                 bool // this is computed from the configuration 2.0
+	CreateCommonTable            bool
+	CreateCommonTableForWildcard bool // the meaning of this is to create a common table for wildcard indexes
```
avelanarius, line 49:
```suggestion
	UseCommonTableForWildcard bool // the meaning of this is to create a common table for wildcard indexes
```

Such naming maybe would make things clearer?

quesma/quesma/config/config_v2.go
```diff
@@ -552,6 +552,9 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 					errAcc = multierror.Append(errAcc, fmt.Errorf("invalid target %s in configuration of %s", target, DefaultWildcardIndexName))
 				}
 			}
+			if defaultConfig.UseCommonTable {
+				conf.CreateCommonTableForWildcard = true
```
avelanarius, line 556: Shouldn't this also set both `conf.CreateCommonTable = true` and `conf.CreateCommonTableForWildcard = true`? (the first one determines if common table is created, the second one controls `table_resolver`)

quesma/quesma/config/config.go
pdelewski: I agree, but should not we change `CreateCommonTable` as well to be consistent?

quesma/quesma/config/config_v2.go
pdelewski: ok, that make sense

pdelewski: However, after consideration I don't think it's needed as there are two separate rules:
- makeDefaultWildcard
- makeCommonTableResolver

avelanarius: What I had in mind was this use of `CreateCommonTable`: https://github.com/QuesmaOrg/quesma/blob/87c168d36135bb99b47c0b77ec1e089aa8f65cb5/quesma/main.go#L105-L108

pdelewski: ok, then you are right

avelanarius (approved):
LGTM, but my personal taste would be to change the name of `CreateCommonTableForWildcard` to `UseCommonTableForWildcard` (or something similar)

pdelewski:
> LGTM, but my personal taste would be to change the name of `CreateCommonTableForWildcard` to `UseCommonTableForWildcard` (or something similar)

ok, I looked at that one more time and I agree as `CreateCommonTable` is used in both contexts for creation and use and the one for wildcard is used only in one context. I will rename it according to your suggestion.

quesma/quesma/config/config.go
pdelewski: done

PR #916, opened 2024-10-24
Fix usage of deprecated `set-output`, give nicer URL to commit

avelanarius:
`set-output` is deprecated (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and the job always printed annoying warning because of that. Fix that to use `$GITHUB_OUTPUT` instead.

Additionally, instead of printing just a hash of the commit, print a (clickable) URL to the commit.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -45,7 +45,8 @@ jobs:
 
       - name: Get last commit author
         id: get_author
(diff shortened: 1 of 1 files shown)


avelanarius:
Tested it on my fork, seemed to work correctly there.

pdelewski (approved):

PR #917, opened 2024-10-24
Using `UseCommonTableForWildcard`property for wildcard config

pdelewski:
This is PR to handle queries  where following configuration is used (no explicit index configuration)

```
'*':
  useCommonTable: true
```

### DIFF
--- a/quesma/quesma/schema_transformer.go
+++ b/quesma/quesma/schema_transformer.go
@@ -16,7 +16,7 @@ import (
 )
 
 type SchemaCheckPass struct {
-	cfg map[string]config.IndexConfiguration
+	cfg *config.QuesmaConfiguration
 }
 
 func (s *SchemaCheckPass) applyBooleanLiteralLowering(index schema.Schema, query *model.Query) (*model.Query, error) {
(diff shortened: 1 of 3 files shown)


quesma/quesma/schema_transformer.go
```diff
@@ -356,6 +356,9 @@ func (s *SchemaCheckPass) applyPhysicalFromExpression(currentSchema schema.Schem
 	if len(query.Indexes) == 1 {
 		if indexConf, ok := s.cfg[query.Indexes[0]]; ok {
 			useCommonTable = indexConf.UseCommonTable
+		} else {
+			// if index configuration is not found, use common table
+			useCommonTable = true
```
avelanarius, line 361: Shouldn't this use `conf.UseCommonTableForWildcard` instead?

pdelewski: :) Just wanted to avoid passing `conf` and open discussion

avelanarius: (outside of scope of this PR, but even better if this code used the `IsCommonTable` from the `table_resolver`'s `Decision` - but this is probably a larger refactor)

pdelewski: ok, I was thinking that this will require more boilerplate, so first wanted to have something working, but it seems that it's easy.

avelanarius (approved):

PR #923, opened 2024-10-28
`bucket_script` (pipeline aggr) enhancements

trzysiek:


quesma/model/pipeline_aggregations/bucket_script.go
```diff
@@ -4,44 +4,85 @@ package pipeline_aggregations
...
-	}
-	var response []model.JsonMap
-	for _, row := range rows {
-		response = append(response, model.JsonMap{"value": row.Cols[0].Value})
-	}
-	return model.JsonMap{
-		"buckets": response,
+	const defaultValue = 0.
+	switch {
+	case query.script == "params.numerator != null && params.denominator != null && params.denominator != 0 ? params.numerator / params.denominator : 0":
```
jakozaur, line 30: Ok, this does not look very generic and is kind of hacky. We can add it for now with some comments that we should implement this fully.

quesma/model/pipeline_aggregations/bucket_script.go
```diff
@@ -4,44 +4,85 @@ package pipeline_aggregations
...
+		query.isCount, query.Parent, query.PathToParent, query.parentBucketAggregation, query.script)
 }
 
 func (query BucketScript) PipelineAggregationType() model.PipelineAggregationType {
-	return model.PipelineParentAggregation // not sure, maybe it's sibling.
+	return model.PipelineParentAggregation // not sure, maybe it's sibling. change hasn't changed the result when running some tests.
+}
+
+func (query BucketScript) findFilterValue(rows []model.QueryResultRow, filterName string) float64 {
+	const defaultValue = 0.0
```
jakozaur, line 71: Does not like this logic, during query generation we should save exact filter name and reuse it here.

quesma/testdata/clients/clover.go
```diff
@@ -0,0 +1,350 @@
...
+				model.NewQueryResultCol("metric__timeseries__61ca57f2-469d-11e7-af02-69e470af7417-denominator_col_0", int64(202)),
+				model.NewQueryResultCol("metric__timeseries__61ca57f2-469d-11e7-af02-69e470af7417-numerator_col_0", int64(0)),
+			}},
+		},
+		ExpectedPancakeSQL: `
+			SELECT count(*) AS "aggr__timeseries__count",
+			  countIf(True) AS
+			  "metric__timeseries__61ca57f2-469d-11e7-af02-69e470af7417-denominator_col_0",
+			  countIf(NOT ("xdr.result_code_str" = 'DIAMETER_SUCCESS')) AS
+			  "metric__timeseries__61ca57f2-469d-11e7-af02-69e470af7417-numerator_col_0"
```
jakozaur, line 345: How this `61ca57f2-469d-11e7-af02-69e470af7417` is generated? Not sure if I recall this logic. I thought it should be just: `metric__timeseries__61ca57f2-469d-11e7-af02-69e470af7417-denominator_col_0`

jakozaur (commented):
Some comments.

quesma/testdata/clients/clover.go
trzysiek: It's a normal aggregation name, results for this tests aren't correct yet.

quesma/model/pipeline_aggregations/bucket_script.go
trzysiek: I don't understand, isn't that exactly what I'm doing? Passing `filterName` as an argument.
I don't save it now, as handling only `numerator/denominator` works

jakozaur (approved):

PR #925, opened 2024-10-29
table_resolver refactor - introduce patternSplitter, decisionMerger

avelanarius:
This PR changes the way patterns (wildcards such as `logs*`) are handled by `table_resolver`. 

Previously, many `table_resolver` rules had two code paths - one code path for a single index (e.g. `logs1`) and one code path for patterns (e.g. `logs1,logs2` or `logs*`), making those rules more complicated. Moreover, one rule `makeCheckIfPatternMatchesAllConnectors` was the primary one to handle patterns and it contained fragments of other rules.

This PR introduces `patternSplitter`, `decisionMerger`. The `table_resolver` now works like this:
1. `patternSplitter` splits a pattern (e.g. `logs*`) into concrete single indexes (e.g. `logs1`, `logs2`)
2. Existing rules are evaluated on each single index separately
3. `decisionMerger` merges decisions of single indexes, making sure that the decisions are compatible. It yields a single decision.

### DIFF
--- a/quesma/table_resolver/rules.go
+++ b/quesma/table_resolver/rules.go
@@ -9,20 +9,71 @@ import (
 	"quesma/end_user_errors"
 	"quesma/quesma/config"
 	"quesma/util"
+	"reflect"
 	"slices"
+	"strings"
 )
 
 // TODO these rules may be incorrect and incomplete
 // They will be fixed int the next iteration.
 
-func patternIsNotAllowed(input parsedPattern) *Decision {
-	if !input.isPattern {
-		return nil
+func (r *tableRegistryImpl) legacyPatternSplitter(pipeline string) func(pattern string) (parsedPattern, *Decision) {
+	return func(pattern string) (parsedPattern, *Decision) {
+		patterns := strings.Split(pattern, ",")
+
+		// Given a (potentially wildcard) pattern, find all non-wildcard index names that match the pattern
+		var matchingSingleNames []string
+		for _, pattern := range patterns {
+			if !elasticsearch.IsIndexPattern(pattern) || elasticsearch.IsInternalIndex(pattern) {
+				matchingSingleNames = append(matchingSingleNames, pattern)
+				continue
+			}
+
+			for indexName := range r.conf.IndexConfig {
+				if util.IndexPatternMatches(pattern, indexName) {
+					matchingSingleNames = append(matchingSingleNames, indexName)
+				}
+			}
+
+			// but maybe we should also check against the actual indexes ??
+			for indexName := range r.elasticIndexes {
+				if util.IndexPatternMatches(pattern, indexName) {
+					matchingSingleNames = append(matchingSingleNames, indexName)
+				}
+			}
+			if r.conf.AutodiscoveryEnabled {
+				for tableName := range r.clickhouseIndexes {
+					if util.IndexPatternMatches(pattern, tableName) {
+						matchingSingleNames = append(matchingSingleNames, tableName)
+					}
+				}
+			}
+		}
+
+		matchingSingleNames = util.Distinct(matchingSingleNames)
+
+		return parsedPattern{
+			source:    pattern,
+			isPattern: len(patterns) > 1 || strings.Contains(pattern, "*"),
+			parts:     matchingSingleNames,
+		}, nil
 	}
-	return &Decision{
-		Reason: "Pattern is not allowed.",
-		Err:    fmt.Errorf("pattern is not allowed"),
+}
+
+func singleIndexSplitter(pattern string) (parsedPattern, *Decision) {
+	patterns := strings.Split(pattern, ",")
+	if len(patterns) > 1 || strings.Contains(pattern, "*") {
+		return parsedPattern{}, &Decision{
+			Reason: "Pattern is not allowed.",
+			Err:    fmt.Errorf("pattern is not allowed"),
+		}
 	}
+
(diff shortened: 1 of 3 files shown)


pdelewski:
I really like logic :)

quesma/table_resolver/table_resolver.go
```diff
@@ -281,8 +292,11 @@ func NewTableResolver(quesmaConf config.QuesmaConfiguration, discovery clickhous
 		pipelineName: IngestPipeline,
 
 		resolver: &compoundResolver{
+			patternSplitter: patternSplitter{
```
pdelewski, line 295: Code is self explaining, however it would make sense to add general idea description (something you added as PR comment)

```
1) patternSplitter splits a pattern (e.g. logs*) into concrete single indexes (e.g. logs1, logs2)
2) Existing rules are evaluated on each single index separately
3) decisionMerger merges decisions of single indexes, making sure that the decisions are compatible. It yields a single decision.
```

quesma/table_resolver/rules.go
```diff
@@ -9,34 +9,81 @@ import (
...
+func (r *tableRegistryImpl) legacyPatternSplitter(pipeline string) func(pattern string) (parsedPattern, *Decision) {
+	return func(pattern string) (parsedPattern, *Decision) {
+		patterns := strings.Split(pattern, ",")
+
+		// Given a (potentially wildcard) pattern, find all non-wildcard index names that match the pattern
+		var matchingSingleNames []string
+		for _, pattern := range patterns {
+			if !elasticsearch.IsIndexPattern(pattern) || elasticsearch.IsInternalIndex(pattern) {
+				matchingSingleNames = append(matchingSingleNames, pattern)
+				continue
```
pdelewski, line 28: Why `continue`?

avelanarius: If `pattern` is not a pattern, but a single index, then there's no need to split it into single indexes (this is what code below does).

If `pattern` is an ElasticSearch internal index then we don't interfere with that (try to split it) - the `resolveInternalElasticName` will immediately decide to route it to Elastic. This is a bit of a workaround...

pdelewski: ok, thx, now it's clear, however maybe short comment would be useful to avoid such questions

quesma/table_resolver/rules.go
```diff
@@ -209,214 +251,168 @@ func (r *tableRegistryImpl) singleIndex(indexConfig map[string]config.IndexConfi
...
-							if index.UseCommonTable {
-								matchedVirtualTables = append(matchedVirtualTables, indexName)
-							} else {
-								matchedTables = append(matchedTables, indexName)
-							}
+			if rhsClickhouse, ok := connDecisionRhs.(*ConnectorDecisionClickhouse); ok {
+				if lhsClickhouse, ok := connDecisionLhs.(*ConnectorDecisionClickhouse); ok {
+					if lhsClickhouse.ClickhouseTableName != rhsClickhouse.ClickhouseTableName {
+						return nil, &Decision{
+							Reason: "Inconsistent ClickHouse table usage",
```
nablaone, line 303: We should add more details here.

quesma/table_resolver/rules.go
```diff
@@ -209,214 +251,168 @@ func (r *tableRegistryImpl) singleIndex(indexConfig map[string]config.IndexConfi
...
+								Reason: "Inconsistent common table usage",
+								Err:    fmt.Errorf("inconsistent common table usage - %s and %s", connDecisionRhs, connDecisionLhs),
+							}
+						}
+						lhsClickhouse.ClickhouseTables = append(lhsClickhouse.ClickhouseTables, rhsClickhouse.ClickhouseTables...)
+						lhsClickhouse.ClickhouseTables = util.Distinct(lhsClickhouse.ClickhouseTables)
+					} else {
+						if !reflect.DeepEqual(lhsClickhouse, rhsClickhouse) {
+							return nil, &Decision{
+								Reason: "Inconsistent ClickHouse table usage",
```
nablaone, line 319: The same reason is returned above. There is no way to distinguish which one rule has been fired.

quesma/table_resolver/rules.go
```diff
@@ -209,214 +251,168 @@ func (r *tableRegistryImpl) singleIndex(indexConfig map[string]config.IndexConfi
...
 							}
 						}
 					}
+					foundMatching = true
 				}
 			}
+		}
+		if !foundMatching {
+			return nil, &Decision{
+				Reason: "Inconsistent connectors",
```
nablaone, line 330: As above, we need more details here.

quesma/table_resolver/rules.go
avelanarius: Fixed in last commit - added comment.

quesma/table_resolver/rules.go
avelanarius: Fixed in last commit - made the reason (and corresponding Err) more clear.

quesma/table_resolver/rules.go
avelanarius: Fixed in last commit - made the reason (and corresponding Err) more clear.

quesma/table_resolver/rules.go
avelanarius: Fixed in last commit - made the reason (and corresponding Err) more clear.

pdelewski (approved):

PR #927, opened 2024-10-29
`Filter(s)` 2nd approach

trzysiek:
So, this isn't a full solution.
Before we only knew how to handle `filter(s)` when they're leaves (or almost) in the aggregation tree.
Now we still can do that, but we also handle when they're the first aggregation layer (we just split them into multiple pancakes, with different `WHERE`)

I find it likely that we won't really need anything more, as those 2 cases exhaust both all our tests, and customer's case, so maybe it's hard to make a dashboard in Kibana with filter in the middle.

But anyway even if we want to support that, adding most top-level filter to where clause seems like a special case and would still be the way to go, so this PR should be a step forward.


jakozaur (commented):
Ok, that's good, we may wish to merge it anyway.

I'm still puzzled with this approach. Way better than previous one, but adding to top where might break total count as well as not work if it's inside. Instead, I would likely keep combinator logic, but just one filter per query each.

jakozaur (approved):
Fine for now.

PR #938, opened 2024-10-31
Make ClickHouse connection configuration more user-friendly

avelanarius:
Before this change, Quesma supported only the "clickhouse://" protocol to connect to ClickHouse. However, ClickHouse also supports the HTTP protocol (8123) and HTTPS protocol (8443). This PR extends the ClickHouse connection logic to switch to using HTTP(S) protocol if the user put that endpoint instead.

This change is motivated by the fact that the HTTPS endpoint is the first/default endpoint shown on ClickHouse Cloud and by the fact that we saw a user try to use the HTTP endpoint.

### DIFF
--- a/quesma/clickhouse/connection.go
+++ b/quesma/clickhouse/connection.go
@@ -16,8 +16,8 @@ import (
 )
 
 func initDBConnection(c *config.QuesmaConfiguration, tlsConfig *tls.Config) *sql.DB {
-
 	options := clickhouse.Options{Addr: []string{c.ClickHouse.Url.Host}}
+
 	if c.ClickHouse.User != "" || c.ClickHouse.Password != "" || c.ClickHouse.Database != "" {
 
 		options.Auth = clickhouse.Auth{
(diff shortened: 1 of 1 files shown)


avelanarius:
@mieciu I vaguely remember you mentioning something about `http://` endpoint in Hydrolix, but from all I've seen Hydrolix also uses the `clickhouse://` endpoint - just making sure that this change is also compatible with Hydrolix (I only tested ClickHouse OSS and ClickHouse Cloud).

mieciu:
> @mieciu I vaguely remember you mentioning something about `http://` endpoint in Hydrolix, but from all I've seen Hydrolix also uses the `clickhouse://` endpoint - just making sure that this change is also compatible with Hydrolix (I only tested ClickHouse OSS and ClickHouse Cloud).

So far we didn't really care about the protocol prefix - once connection has been established with host we relied on driver defaults, so that'd be something to check. I'm not sure if querying over http protocol is supported there.

quesma/quesma/config/url.go
```diff
@@ -15,6 +18,12 @@ func (u *Url) UnmarshalText(text []byte) error {
 	if err != nil {
 		return err
 	}
+	if len(urlValue.Scheme) == 0 {
+		return fmt.Errorf("URL scheme (e.g. http:// or clickhouse://) is missing from the provided URL: %s", urlValue)
+	}
+	if len(urlValue.Port()) == 0 {
+		return fmt.Errorf("URL port (e.g. 8123 in 'http://localhost:8123') is missing from the provided URL: %s", urlValue)
```
mieciu, line 25: This URL parsing in Go is borked - below you can see the result of parsing `demo.host.com:9440`:
image
which will result in `URL port is missing`.

I don't like suggesting this, but if we want to get correct messages we should introduce our own little parsing logic (e.g. extracting `.*://` or `:[[:digit:]]`)  😬

avelanarius: In order to avoid introducing our own parsing code, I changed the logic from: `len(urlValue.Scheme) == 0` to `len(urlValue.Scheme) == 0 || len(urlValue.Host) == 0` and made the error message more generic: `"Invalid URL (missing protocol and/or hostname). Expected URL in a format 'protocol://hostname:port' (e.g. 'http://localhost:8123'), but instead got: %s"`

avelanarius:
Since there are some quirks with HTTP(S) in Hydrolix and this PR might not be mergable as-is, I extracted only the validation improvements to this (mergable) PR  https://github.com/QuesmaOrg/quesma/pull/946

avelanarius:
And the quirks are:
- Hydrolix native protocol endpoint uses the `https://` scheme: https://docs.hydrolix.io/docs/query-some-data
  > Connect to Hydrolix
  > HTTPS API: https://{myhostname}}.hydrolix.live/query
  > Clickhouse **Native: https://**{myhostname}}.hydrolix.live:9440
- Hydrolix HTTPS API has a different API than ClickHouse HTTP(S) API: `curl --request POST \
     --url https://hostname.hydrolix.live/query \
     --header 'accept: text/json' \
     --header 'content-type: text/plain' \
     --data 'SELECT count() FROM sample.taxi_trips'` vs `curl 'http://localhost:8123/?query=SELECT%201'`

mieciu:
Yea just to leave trace after our investigation:
* http to ClickHouse works because we rely on https://github.com/ClickHouse/clickhouse-go library
* http to Hydrolix currently is not working
This discrepancy is related to differences in how queries over http are handled in both DBs. 

Relevant sources: [ClickHouse](https://clickhouse.com/docs/en/interfaces/http) / [Hydrolix](https://docs.hydrolix.io/reference/query-data-post).

pdelewski:
I'm closing it due to mentioned quirks above and divergence from main branch.

PR #939, opened 2024-10-31
Minor fixes (some TODOs)

pdelewski:
Minor improvements, removing dead code

### DIFF
--- a/quesma/clickhouse/parser.go
+++ b/quesma/clickhouse/parser.go
@@ -113,26 +113,8 @@ func RemoveNonSchemaFields(m SchemaMap, t *Table) SchemaMap {
 				delete(mCur, fieldName)
 			}
 		}
-		/*
-			for fieldName, v := range mCur {
-				mSchemaNested, ok := col.[fieldName]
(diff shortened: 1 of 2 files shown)


quesma/clickhouse/parser.go
```diff
@@ -141,9 +123,12 @@ func RemoveNonSchemaFields(m SchemaMap, t *Table) SchemaMap {
 				descendRec(col, vCasted)
 			}
 		} else {
-			delete(m, fieldName) // TODO check if it's fine? In c++ not, but here seems to work
+			toDelete = append(toDelete, fieldName)
```
avelanarius, line 126: Hmm, the internet says it's OK to delete elements from map while iterating over it? (so the previous, simpler approach)

https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

pdelewski: That's interesting, but I was basing on other post that was telling that's unsafe

PR #941, opened 2024-11-01
"fix"/hack

trzysiek:
Ideally I wouldn't like to merge any of this, maybe besides size being string, but I'm afraid most or all of this was needed for all dashboards to work. I'll need to investigate.


quesma/queryparser/query_parser.go
```diff
@@ -1262,6 +1262,13 @@ func (cw *ClickhouseQueryTranslator) parseSize(queryMap QueryMap, defaultSize in
 		return defaultSize
 	} else if sizeAsFloat, ok := sizeRaw.(float64); ok {
 		return int(sizeAsFloat)
+	} else if sizeAsString, ok := sizeRaw.(string); ok {
```
mieciu, line 1265: I'd suggest moving this to separate PR and merging to `main`

mieciu: By "this" I mean the whole `quesma/queryparser/query_parser.go` change of course 😉

trzysiek:
No need for it, it even stopped passing tests.

PR #945, opened 2024-11-04
Target config extended configuration

pdelewski:
This PR implements new variant of `target` configuration where it's a list not an array and can have more attributes (potentially in the future)

Current variant is still supported and is a fallback in the case extended version is missing

Below
```
  kibana_sample_data_flights:
    target:
    - my-clickhouse-data-source:
        useCommonTable: true
```
is equivalent of 
```
  kibana_sample_data_flights:
    target: [my-clickhouse-data-source]
    useCommonTable: true
```

PR contains a bit of duplicated code, something that I would prefer fix in next PR.

### DIFF
--- a/quesma/quesma/config/config_v2.go
+++ b/quesma/quesma/config/config_v2.go
@@ -575,9 +575,9 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 					} else {
 						errAcc = multierror.Append(errAcc, fmt.Errorf("invalid target %s in configuration of %s", target, DefaultWildcardIndexName))
 					}
-					if _, exists := target.properties["useCommonTable"]; exists {
-						conf.CreateCommonTable = true
-						conf.UseCommonTableForWildcard = true
+					if val, exists := target.properties["useCommonTable"]; exists {
+						conf.CreateCommonTable = val == "true"
+						conf.UseCommonTableForWildcard = val == "true"
 					}
 				}
 			}
@@ -622,8 +622,8 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 						} else {
 							errAcc = multierror.Append(errAcc, fmt.Errorf("invalid target %s in configuration of index %s", target, indexName))
 						}
-						if _, exists := target.properties["useCommonTable"]; exists {
-							processedConfig.UseCommonTable = true
+						if val, exists := target.properties["useCommonTable"]; exists {
+							processedConfig.UseCommonTable = val == "true"
 						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


quesma/quesma/config/test_configs/target_new_variant.yaml
```diff
@@ -0,0 +1,71 @@
...
+    type: quesma-v1-processor-query
+    config:
+      indexes:
+        kibana_sample_data_ecommerce:
+          target:
+            - my-clickhouse-data-source:
+                useCommonTable: false
+        kibana_sample_data_flights:
+          target:
+            - my-clickhouse-data-source: {}
```
avelanarius, line 38: Is there some way to avoid having to write `{}`?

pdelewski: I don't know. It seems to be yaml syntax specific

quesma/quesma/config/config_v2.go
```diff
@@ -709,14 +833,34 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 			}
 
 			processedConfig.IngestTarget = make([]string, 0) // reset previously set defaultConfig.IngestTarget
-			for _, target := range indexConfig.Target {
-				if targetType, found := c.getTargetType(target); found {
-					processedConfig.IngestTarget = append(processedConfig.IngestTarget, targetType)
-				} else {
-					errAcc = multierror.Append(errAcc, fmt.Errorf("invalid target %s in configuration of index %s", target, indexName))
+			targets, errTarget = c.getTargetsExtendedConfig(indexConfig.Target)
```
avelanarius, line 836: This `if len(targets) > 0`, `if len(targets) == 0` pattern repeats a few times in the code. Maybe this logic should be moved into `c.getTargetsExtendedConfig`?

What I mean by that is that if simplified target configuration is used, then `getTargetsExtendedConfig` would "translate" simplified target configuration into extended configuration: translate `[target1, target2]` into array of structs `[{target: target1, properties: {}}, {target: target2, properties: {}}]`

pdelewski: Yes, we can do that, but as I mentioned. I would prefer to do that in next PR if you don't mind, to have working version first

avelanarius (approved):
The new configuration syntax looks OK, but I think this looks strange:
```yaml
        kibana_sample_data_flights:
          target:
            - my-clickhouse-data-source: {}
```
and it'd be great if something like this was possible instead:
```yaml
        kibana_sample_data_flights:
          target:
            - my-clickhouse-data-source
```

pdelewski:
> The new configuration syntax looks OK, but I think this looks strange:
> 
> ```yaml
>         kibana_sample_data_flights:
>           target:
>             - my-clickhouse-data-source: {}
> ```
> 
> and it'd be great if something like this was possible instead:
> 
> ```yaml
>         kibana_sample_data_flights:
>           target:
>             - my-clickhouse-data-source
> ```

Let me check that again

PR #946, opened 2024-11-04
Improve connection doctor and validation of connection URLs

avelanarius:
This PR improves the validation of connection URLs in configuration, giving clearer error messages to the user and the expected format of the connection string.

Related to that, RemoteLogDrainUrl and QuesmaInternalTelemetryUrl now can't be provided in the config - they were overwritten by hardcoded constants either way.

Additionally, ClickHouse connection doctor is improved:
- If connection succeeded it suggests making sure that username/password is correct
- It now skips TLS trial connections if the user disabled TLS in the config (and informs of that fact)
- It tries default ports in more cases

This PR is extracted from larger PR #938.


mieciu (approved):
👍

PR #951, opened 2024-11-05
Extended target config improvements, reducing code redundancy

pdelewski:


quesma/quesma/config/config_v2.go
```diff
@@ -979,6 +883,14 @@ func (c *QuesmaNewConfiguration) getTargetsExtendedConfig(target any) ([]struct
...
+			if targetName, ok := target.(string); ok {
+				result = append(result, struct {
+					target     string
+					properties map[string]interface{}
+				}{target: targetName, properties: map[string]interface{}{}})
+			}
+		}
 		for _, target := range targets {
 			if targetMap, ok := target.(map[string]interface{}); ok {
 				for name, settings := range targetMap {
```
avelanarius, line 896: I am worried that this messes up the order of targets in case some targets are `string`s and some are `map[string]interface{}` (not sure if it's possible). And the order of targets matters (for A/B testing for example).

You could easily solve this problem by moving those two ifs into a single loop, not two loops.

pdelewski: ok, I was just thinking that this will be cleaner to have two passes, but it seems that your suggestion does not break tests.

avelanarius (approved):

PR #956, opened 2024-11-06
Remove string specific array transfomer

nablaone:
We can remove that `string` specific transformer. We've got a generic solution already.

Before:

Clickhouse on the left. 

before

dashboard-before

After:

Clickhouse on the left. 
after
dashboard-after


mieciu (approved):
image

PR #959, opened 2024-11-07
Improve login redirect

jakozaur:
@mieciu please test it before merging. I could not do that locally.

It fixes two issues:
- right redirect if you host it on different domain than localhost
- do also redirect if HTMX hit it in some fragment (auto-refreshed panel)


mieciu (approved):
✅ 
(I also spotted one trivial thing which I've just committed)

PR #967, opened 2024-11-11
A/B - Performance gain change

nablaone:
This PR adds a different rendering of performance gain for extreme values: 
1. when both response times are speedy (<50ms) we don't compute a gain, it doesn't make sense  
2. when one of the sides is significantly slower (over 5s), we show which is slower (it may be an infrastructure issue)

Screenshot 2024-11-11 at 14 51 48


nablaone:
We should replace "Performance gain" with "Response time delta".

We should compute how much clickhouse is faster than elastic. Elastic time is our baseline.

quesma/quesma/ui/html_utils.go
```diff
@@ -81,9 +81,9 @@ func (qmc *QuesmaManagementConsole) generateTopNavigation(target string) []byte
 	if target == "ab-testing-dashboard" {
 		buffer.Html(` class="active"`)
 	}
-	buffer.Html(`><a href="`)
+	buffer.Html(`><a title="Compatibility Report" href="`)
 	buffer.Html(abTestingPath)
-	buffer.Html(`">A/B</a></li>`)
+	buffer.Html(`">CR</a></li>`)
```
avelanarius, line 86: Don't forget that there are instance of "A/B testing" still left in `ab_testing.go` (such as `A/B Testing - Panel Details`).

It's OK if you want to do it in a followup PR.

quesma/quesma/ui/ab_testing.go
```diff
@@ -405,8 +428,32 @@ func (qmc *QuesmaManagementConsole) generateABTestingReport(kibanaUrl, orderBy s
...
+
+			case *row.aTime > maxTime && *row.bTime < maxTime:
+				buffer.Text(fmt.Sprintf("%s is over %0.02fs", row.aName, *row.aTime))
+
+			case *row.bTime > maxTime && *row.aTime < maxTime:
+				buffer.Text(fmt.Sprintf("%s is over %0.02fs", row.bName, *row.bTime))
+
+			default:
+
+				buffer.Text(fmt.Sprintf("%.01f%%", *row.responseTimeDelta))
```
avelanarius, line 453:
```suggestion
                                if *row.responseTimeDelta > 0 {
				    buffer.Text(fmt.Sprintf("%.01f%% faster", *row.responseTimeDelta))
				} else {
				    buffer.Text(fmt.Sprintf("%.01f%% slower", -(*row.responseTimeDelta)))
				}
```

avelanarius: if we want to be fancier

avelanarius (approved):
LGTM, left 2 suggestions

avelanarius:
gofmt fails (tabs vs spaces 😢)

Issue #944, opened 2024-11-01
Remove filters limitations


jakozaur:
We allow `filters` only as top or as leaf bucket aggregations. We also might return undercount if we have top-level `filters`. 

We should allow them everywhere.

See:
https://github.com/QuesmaOrg/quesma/pull/927

PR #970, opened 2024-11-11
Improve `(date)_range` just like `filters`

trzysiek:
`(date)_range` and `filters` is basically the same aggregation. I improved `filters` a bit recently, but didn't do anything to `(date)_range`.
This PR makes all those combinators work in the same way, so there's one issue left -  https://github.com/QuesmaOrg/quesma/issues/944 Afterwards they will all fully work.


quesma/queryparser/pancake_sql_query_generation_test.go
```diff
@@ -52,8 +52,9 @@ func TestPancakeQueryGeneration(t *testing.T) {
 			if filters(test.TestName) {
 				t.Skip("Fix filters")
 			}
-			if test.TestName == "complex sum_bucket. Reproduce: Visualize -> Vertical Bar: Metrics: Sum Bucket (Bucket: Date Histogram, Metric: Average), Buckets: X-Asis: Histogram(file:opensearch-visualize/pipeline_agg_req,nr:22)" {
-				t.Skip("error: filter(s)/range/dataRange aggregation must be the last bucket aggregation")
+
+			if test.TestName == "Line, Y-axis: Min, Buckets: Date Range, X-Axis: Terms, Split Chart: Date Histogram(file:kibana-visualize/agg_req,nr:9)" {
+				t.Skip("Date range is broken, fix in progress (PR #971)")
```
trzysiek, line 57: Adding a new test that fails. Unskipping another one above/below.

quesma/queryparser/pancake_transformer.go
```diff
@@ -408,48 +408,61 @@ func (a *pancakeTransformer) aggregationTreeToPancakes(topLevel pancakeAggregati
...
+	if !isCombinator {
+		return
+	}
+
+	noMoreBucket := len(pancake.layers) <= 1 || (len(pancake.layers) == 2 && pancake.layers[1].nextBucketAggregation == nil)
+	noMetricOnFirstLayer := len(firstLayer.currentMetricAggregations) == 0 && len(firstLayer.currentPipelineAggregations) == 0
+	canSimplyAddCombinatorToWhereClause := noMoreBucket && noMetricOnFirstLayer
+	if canSimplyAddCombinatorToWhereClause {
+		return
+	}
```
trzysiek, line 443: I think code should be a bit clearer now? Feel free to disagree

jakozaur: Though not sure if that's always correct if we count size. Though it might be.

jakozaur: Ok, we add size as metric aggregation, so this should be good.

jakozaur (approved):
Great work!

PR #976, opened 2024-11-12
Combine schema and JSON types for Array fields

avelanarius:
When creating a table, Quesma uses both the schema information (which can come from "PUT /:index/_mapping" or YAML configuration) and the inferred types from JSON. However, the schema information does not contain knowledge whether a field is an array or not - in ElasticSearch essentially every field can become an array (a field with type "keyword" can receive an array of keywords).

Therefore Quesma has to handle arrays in a different way. Previously if it detected an array, it would not use schema information at all. This PR improves this logic - schema information is now combined with the JSON information (that a field is actually an array).


avelanarius:
Extracted a part of it to https://github.com/QuesmaOrg/quesma/pull/1248

pdelewski:
Closing this due to conflicts and fork. Continuation in https://github.com/QuesmaOrg/quesma/pull/1427

PR #981, opened 2024-11-13
Disable logging of SQL queries 

nablaone:
Logging every SQL query that is executed or transformer has some performance penalty. This PR adds a flag to enable it. We should not log queries by default. 

Tracing can be enabled via conf:
```
logging:
  enableSQLTracing: true
```


pdelewski:
LGTM

pdelewski (approved):

PR #982, opened 2024-11-13
Handle `/_async_search/status/:id` endpoint

mieciu:
Initially this meant to be just addition of an endpoint, but eventually it exposed quite serious bug in `quesma/quesma/router.go`.

### 1. Adding `/_async_search/status/:id` handling 

Before that change, requests targeting this endpoint were routed to Elasticsearch which of course had no recollection of our own async IDs `quesma_async_019329bf-4330-7007-970d-e87734bca7c3`. 

This fixes the following behavior, esp. visible when querying larger data sets:
image

### 2. Fixing `router.go`

The root cause of this bug was in `findHandler` function of `quesma/mux/mux.go`. Basically, whenever we registered same route twice with diffrent matchers (e.g. for two separate HTTP methods), `findHandler` always ditched the second registered func. See screenshot below:
image

Of course I had to add a pretty large test to make sure I don't break stuff.

Closes: https://github.com/QuesmaOrg/quesma/issues/995

### DIFF
--- a/quesma/quesma/mux/mux.go
+++ b/quesma/quesma/mux/mux.go
@@ -99,10 +99,8 @@ func (p *PathRouter) Matches(req *Request) (Handler, *table_resolver.Decision) {
 	}
 }
 
-func (p *PathRouter) findHandler(req *Request) (Handler, *table_resolver.Decision) {
+func (p *PathRouter) findHandler(req *Request) (handler Handler, decision *table_resolver.Decision) {
 	path := strings.TrimSuffix(req.Path, "/")
-	var handler Handler
-	var decision *table_resolver.Decision
 	for _, m := range p.mappings {
 		if pathData, pathMatches := m.compiledPath.Match(path); pathMatches {
 			req.Params = pathData.Params
@@ -113,10 +111,7 @@ func (p *PathRouter) findHandler(req *Request) (Handler, *table_resolver.Decisio
 			}
 		}
 	}
-	if handler != nil {
-		return handler, decision
-	}
-	return nil, nil
+	return handler, decision
 }
 
 type httpMethodPredicate struct {


quesma/quesma/mux/mux.go
```diff
@@ -86,36 +86,37 @@ func (p *PathRouter) Register(pattern string, predicate RequestMatcher, handler
...
-				return m.handler, true, predicateResult.Decision
-			} else {
-				return nil, false, predicateResult.Decision
+				handler = m.handler
+				decision = predicateResult.Decision
 			}
 		}
 	}
-	return nil, false, nil
+	if handler != nil {
```
pdelewski, line 116: This `if` seems to be redundant

quesma/quesma/router_test.go
```diff
@@ -263,3 +270,112 @@ func Test_matchedAgainstBulkBody(t *testing.T) {
...
+	tr := TestTableResolver{}
+	testRouter := ConfigureRouter(cfg, schema.NewSchemaRegistry(fixedTableProvider{}, cfg, clickhouse.SchemaTypeAdapter{}), &clickhouse.LogManager{}, &ingest.IngestProcessor{}, &ui.QuesmaManagementConsole{}, telemetry.NewPhoneHomeAgent(cfg, nil, ""), &QueryRunner{}, tr)
+
+	tests := []struct {
+		path                string
+		method              string
+		shouldReturnHandler bool
+	}{
+		// Routes explicitly registered in the router code
+		{routes.ClusterHealthPath, "GET", true},
```
pdelewski, line 293: I'm wondering if we could get these paths from router without explicitly listing them here.

pdelewski:
LGTM :)

pdelewski (approved):

quesma/quesma/router_test.go
mieciu: I thought about that, but then the nature of this failure made me think that we'd be safer this way 😉

Issue #968, opened 2024-11-11
`composite` aggregation is not supported


nablaone:
When we try to execute a query with the `composite` aggregation, the empty result is returned. 

How to reproduce. 
1. Call Quesma
```
POST localhost:8080/kibana_sample_data_logs/_search
Content-Type: application/json

POST localhost:8080/kibana_sample_data_logs/_search
Content-Type: application/json

{
  "aggs": {
    "table": {
      "composite": {
        "size": 10000,
        "sources": [
          {
            "stk1": {
              "terms": {
                "field": "machine.os.keyword"
              }
            }
          },
          {
            "stk2": {
              "terms": {
                "field": "geo.dest"
              }
            }
          }
        ]
      }
    }
  },
  "query": {

  },
  "size": 0,
  "track_total_hits": false
}
```
2. Quesma returns an empty response
3. There is a message logged
```
quesma-1  | Nov 11 14:00:50.182 ERR quesma/queryparser/pancake_aggregation_parser.go:185 > unexpected type of subaggregation: (composite: map[size:10000 sources:[map[stk1:map[terms:map[field:machine.os.keyword]]] map[stk2:map[terms:map[field:geo.dest]]]]]), value type: map[string]interface {}. Skipping opaque_id= path=/kibana_sample_data_logs/_search reason="unsupported_search_query: composite" request_id=01931b86-f704-71b9-98e1-c431ee0c039d
```

jakozaur:
We added support for `Composite` aggregation.

PR #1005, opened 2024-11-15
Add new `Composite` aggregation

trzysiek:
https://github.com/QuesmaOrg/quesma/issues/968


quesma/model/README.md
```diff
@@ -15,7 +15,7 @@ More info: https://www.elastic.co/guide/en/elasticsearch/reference/current/searc
  Cardinality               |   :white_check_mark:   | Auto-interval date histogram |    :wavy_dash:     | Bucket script          |     :wavy_dash:    |
  Extended Stats            | :white_check_mark:[^1] | Categorize text              |        :x:         | Bucket count K-S test  |        :x:         |
  Avg                       |   :white_check_mark:   | Children                     |        :x:         | Bucket correlation     |        :x:         |
- Boxplot                   |          :x:           | Composite                    |        :x:         | Bucket selector        |        :x:         |
+ Boxplot                   |          :x:           | Composite                    | :white_check_mark: | Bucket selector        |        :x:         |
```
avelanarius, line 18: Unrelated to this PR, but maybe we should move this table to our public-facing docs?

trzysiek: I think so too!

jakozaur: It does not look like composite works now with arrays as they described. Though still major step forward.

jakozaur (approved):

PR #1006, opened 2024-11-15
Report errors in queries better #2 (in parsing bucket aggregations)

trzysiek:
**old comment (motivation):**
Our error messages for errors encountered while parsing  query request isn't as good as Elastic's, I'll try to improve it here. (It's not that unimportant, with invalid query request where Elastic always fails with an error response, we often don't and e.g. return empty results, which might trick the user that queries went fine)

I think I'll split it into few PRs. This might be a good starting point for another one. (update: actually similar small improvement was already merged, but for pipeline aggregations)

**new comment (what's been done):**
There's really not much going on here.
Clue of this PR is to change our very ugly 400-line `func (cw *ClickhouseQueryTranslator) pancakeTryBucketAggregation` function into something a bit nicer. And also report errors in more places, to be consistent with Elastic's behaviour.

Quite a few style improvements that I automatically caught in the process, but only some trivial/very local ones, so extracting copy/paste used 5 times into 1 function, consistent naming across parsers of different aggregations, etc. (not much more).


quesma/model/bucket_aggregations/dateRange.go
trzysiek, line 1: only very very minor style improvements in this file

quesma/queryparser/aggregation_date_range_parser.go
```diff
@@ -3,80 +3,53 @@
...
-	}
-	var ranges []any
-	var ok bool
-	if formatRaw, exists := dateRange["format"]; exists {
-		if formatParsed, ok := formatRaw.(string); ok {
-			format = formatParsed
-		} else {
-			logger.WarnWithCtx(cw.Ctx).Msgf("format specified for date range aggregation is not a string. Using empty. Querymap: %v", dateRange)
-		}
+func (cw *ClickhouseQueryTranslator) parseDateRangeAggregation(aggregation *pancakeAggregationTreeNode, params QueryMap) (err error) {
```
jakozaur, line 11: I wonder about type signature? Why pass `aggregation` as argument, shouldn't we just return optional queryType?

quesma/queryparser/filters_aggregation.go
```diff
@@ -3,41 +3,27 @@
...
 	"quesma/model"
 	"quesma/model/bucket_aggregations"
+	"sort"
 )
 
-func (cw *ClickhouseQueryTranslator) parseFilters(queryMap QueryMap) (success bool, filtersAggr bucket_aggregations.Filters) {
-	filtersAggr = bucket_aggregations.NewFiltersEmpty(cw.Ctx)
-
-	filtersRaw, exists := queryMap["filters"]
+func (cw *ClickhouseQueryTranslator) parseFilters(aggregation *pancakeAggregationTreeNode, params QueryMap) error {
```
jakozaur, line 12: Similar, here. Not sure if it's right to pass `aggregation` here.

quesma/queryparser/pancake_aggregation_parser_buckets.go
```diff
@@ -5,378 +5,310 @@ package queryparser
...
-func (cw *ClickhouseQueryTranslator) pancakeTryBucketAggregation(aggregation *pancakeAggregationTreeNode, queryMap QueryMap) (success bool, err error) {
-
-	success = true // returned in most cases
-	if histogramRaw, ok := queryMap["histogram"]; ok {
-		histogram, ok := histogramRaw.(QueryMap)
-		if !ok {
-			return false, fmt.Errorf("histogram is not a map, but %T, value: %v", histogramRaw, histogramRaw)
-		}
+func (cw *ClickhouseQueryTranslator) pancakeTryBucketAggregation(aggregation *pancakeAggregationTreeNode, queryMap QueryMap) error {
+	aggregationHandlers := []struct {
```
jakozaur, line 18: Super nice!

quesma/queryparser/aggregation_date_range_parser.go
jakozaur: Ok, after reading later it become obvious why you did that way.

quesma/queryparser/filters_aggregation.go
jakozaur: Ok, I understand. It explain later. Keep as ai.

jakozaur (approved):

PR #1008, opened 2024-11-15
Always return decision in `findHandler`

avelanarius:
d36b617 changed the logic in `findHandler` when it comes to returning decision: after d36b617, the decision is only filled out if the predicate matched (otherwise it's `nil`).

Previously, the decision would be returned in both scenarios: if predicate matched or if predicate did not match.

d36b617 caused a regression in integration tests: if an index is closed in the configuration, the predicate does not match, but the code still needs a decision (with `IsClosed: true`) to be able to return a correct `index_closed_exception` response. Without a decision (`nil`) it wasn't able to return that expected response.


avelanarius:
Integration tests pass with this PR: https://github.com/QuesmaOrg/quesma/actions/runs/11860557379/job/33056109112

trzysiek (approved):

PR #1013, opened 2024-11-18
Router fixes 

nablaone:
PR #982 changed how the router works. It fixes some issues, but introduces a bug: "PUT _bulk"  is treated as "PUT :index".

In this PR :
1. Revert router changes (#982)
2. Fix async handling 
3. Fix mapping handling 
4. Fix "GET :index" handling  (it was always redirected to the elastic)
5. Handle Kibana internal indexes in an explicit way

### DIFF
--- a/quesma/elasticsearch/index.go
+++ b/quesma/elasticsearch/index.go
@@ -15,5 +15,5 @@ func IsIndexPattern(index string) bool {
 }
 
 func IsInternalIndex(index string) bool {
-	return strings.HasPrefix(index, internalIndexPrefix)
+	return strings.HasPrefix(index, internalIndexPrefix) || index == "_nodes" || index == "_xpack"
 }
--- a/quesma/elasticsearch/index.go
+++ b/quesma/elasticsearch/index.go
@@ -15,5 +15,8 @@ func IsIndexPattern(index string) bool {
 }
 
 func IsInternalIndex(index string) bool {
-	return strings.HasPrefix(index, internalIndexPrefix) || index == "_nodes" || index == "_xpack"
+	return strings.HasPrefix(index, internalIndexPrefix)
 }
+
+// InternalPaths is a list of paths that are considered internal and should not handled by Quesma
+var InternalPaths = []string{"/_nodes", "/_xpack"}
--- a/quesma/quesma/mux/mux.go
+++ b/quesma/quesma/mux/mux.go
@@ -86,20 +86,20 @@ func (p *PathRouter) Register(pattern string, predicate RequestMatcher, handler 
 
 }
 
-func (p *PathRouter) Matches(req *Request) (Handler, bool, *table_resolver.Decision) {
-	handler, found, decision := p.findHandler(req)
-	if found {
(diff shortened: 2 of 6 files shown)


quesma/elasticsearch/index.go
```diff
@@ -15,5 +15,5 @@ func IsIndexPattern(index string) bool {
 }
 
 func IsInternalIndex(index string) bool {
```
mieciu, line 17: Just for precision/clarity, I'd add one more func `IsInternalElasticsearchPath` -> `_nodes` and `_xpack` are internal endpoints, not indices

quesma/quesma/mux/mux.go
```diff
@@ -86,32 +86,34 @@ func (p *PathRouter) Register(pattern string, predicate RequestMatcher, handler
 
 }
 
-func (p *PathRouter) Matches(req *Request) (Handler, *table_resolver.Decision) {
-	handler, decision := p.findHandler(req)
-	if handler != nil {
+func (p *PathRouter) Matches(req *Request) (Handler, bool, *table_resolver.Decision) {
```
mieciu, line 89: I don't think that we really need that `bool` in function signature 🧐

quesma/quesma/mux/mux.go
```diff
@@ -86,32 +86,34 @@ func (p *PathRouter) Register(pattern string, predicate RequestMatcher, handler
...
 	} else {
 		routerStatistics.addUnmatched(req.Path)
 		logger.Debug().Msgf("Non-matched path: %s", req.Path)
-		return handler, decision
+		return handler, false, decision
 	}
 }
 
-func (p *PathRouter) findHandler(req *Request) (handler Handler, decision *table_resolver.Decision) {
+func (p *PathRouter) findHandler(req *Request) (Handler, bool, *table_resolver.Decision) {
```
mieciu, line 102: Same as with matches, I'd prefer to remove that bool

quesma/quesma/router.go
```diff
@@ -200,33 +200,38 @@ func ConfigureRouter(cfg *config.QuesmaConfiguration, sr schema.Registry, lm *cl
 		return elasticsearchQueryResult(string(responseBody), http.StatusOK), nil
 	})
 
-	router.Register(routes.IndexMappingPath, and(method("PUT"), matchedAgainstPattern(tableResolver)), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
-		index := req.Params["index"]
+	router.Register(routes.IndexMappingPath, and(method("GET", "PUT"), matchedAgainstPattern(tableResolver)), func(ctx context.Context, req *mux.Request) (*mux.Result, error) {
```
mieciu, line 203: We should be really vocal about ConfigureRouter's behavior w.r.t. handler registering and matching, please add a comment about its behavior (preferably above the function signature so that's its visible in IDE)

mieciu (changes requested):
Most importantly, I'd prefer to avoid that additional `boolean` flying around. 
I liked the concise `if handler != nil then handle`

PR #1019, opened 2024-11-20
Fix `SchemaToColumns` for `geo_point` types, add ingest integration tests

avelanarius:
When a `geo_point` field is added to `fieldEncodings`, actually two fields get added:
- `my_field.lat`
- `my_field.lon`

`SchemaToColumns` however tried to find `my_field` in `fieldEncodings` (which was missing; only `my_field.lat` and `my_field.lon` were there). This resulted in creating (incorrect, unused) columns `lat`, `lon` since `internalPropertyName` was empty.

Fix the problem by accessing `my_field.lat` and `my_field.lon` keys in `fieldEncodings` instead (which should be there).

Add new integration test that checks Quesma's ingest of Kibana sample flights, Kibana sample ecommerce. The tests without "PUT mapping" caught the fixed issue (this is how I found it).

### DIFF
--- a/ci/it/configs/quesma-ingest.yml.template
+++ b/ci/it/configs/quesma-ingest.yml.template
@@ -1,0 +1,142 @@
+frontendConnectors:
+  - name: elastic-ingest
+    type: elasticsearch-fe-ingest
+    config:
+      listenPort: 8080
+  - name: elastic-query
+    type: elasticsearch-fe-query
+    config:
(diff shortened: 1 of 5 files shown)


avelanarius:
Integration tests passed: https://github.com/QuesmaOrg/quesma/actions/runs/11937331599

pdelewski (approved):

Issue #1017, opened 2024-11-20
Cleanup small modules and circular dependencies 


mieciu:
Before we even think about bigger refactoring and bold architectural changes, a tiny prep work today will make those challenges much easier in the future.

We have good amount of very small modules which is going to make our big refactor more difficult.

mieciu:
Enough cleaning for now.

PR #1021, opened 2024-11-21
Remove `quesma/logger` from `utils` package

mieciu:
A step towards untangling dependency hell with our modules.

Mainly it's about the having functions from `utils` return an `error` and logging them (or sometimes not - whenever I've though it's not that significant) in the places they're called.

Related to https://github.com/QuesmaOrg/quesma/issues/1017


quesma/queryparser/pancake_json_rendering.go
```diff
@@ -179,7 +179,11 @@ func (p *pancakeJSONRenderer) combinatorBucketToJSON(remainingLayers []*pancakeM
 		if err != nil {
 			return nil, err
 		}
-		return util.MergeMaps(p.ctx, aggJson, subAggr), nil
+		mergeResult, mergeErr := util.MergeMaps(aggJson, subAggr)
+		if mergeErr != nil {
```
pdelewski, line 183: I'm wondering if we could propagate this error further without logging it here

pdelewski (approved):

quesma/queryparser/pancake_json_rendering.go
mieciu: That'd affect the sub aggregations, I didn't want to touch that at this moment :)

PR #1029, opened 2024-11-22
Remove unnecessary AST serializations

nablaone:


pdelewski:
@nablaone This is about logging according to flag, right?

nablaone:
> @nablaone This is about logging according to flag, right?

Yes.

pdelewski (approved):

PR #1030, opened 2024-11-22
Support auth against OpenSearch

mieciu:
Little hackish, as it involves additional call to the backend (in order to decide whether use OpenSearch or Elasticsearch security endpoint) **but** given that we have auth header cache I think we might get away with that 😉 

FWIW,  Elasticsearch 7 uses the same [endpoint (`_security/_authenticate`)](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-authenticate.html) as Elasticsearch 8 for auth (just verified :relieved:)

### DIFF
--- a/quesma/elasticsearch/client.go
+++ b/quesma/elasticsearch/client.go
@@ -10,10 +10,15 @@ import (
 	"net/http"
 	"quesma/logger"
 	"quesma/quesma/config"
+	"quesma/util"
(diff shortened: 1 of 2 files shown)


nablaone (approved):
LGTM

avelanarius (approved):

pdelewski (approved):

PR #1039, opened 2024-11-25
Usage report for clickhouse

mieciu:


quesma/telemetry/phone_home.go
```diff
@@ -539,6 +596,11 @@ func (a *agent) collect(ctx context.Context, reportType string) (stats PhoneHome
 	} else {
 		stats.ClickHouse = ClickHouseStats{Status: "paused"}
 	}
+	if !strings.HasPrefix(a.config.ClickHouse.ConnectorType, "hydrolix") { // we only check table sizes for ClickHouse
+		if tables, err := a.collectClickHouseTableSizes(ctx); err == nil {
+			logger.Info().Msgf("Usage report dababase=[%s] table sizes=%v", a.getDbInfo(), tables)
```
nablaone, line 601: Why is this logged only?  We can add this site to the clickhouse stats.

quesma/telemetry/phone_home.go
```diff
@@ -296,6 +296,63 @@ where active
...
+func (a *agent) collectClickHouseTableSizes(ctx context.Context) (map[string]int64, error) {
+	tablesWithSizes, err := a.getTableSizes(a.ctx)
+	if err != nil {
+		logger.WarnWithCtx(ctx).Msgf("Error getting table sizes from clickhouse: %v", err)
+		return nil, err
+	}
+	return tablesWithSizes, nil
+}
+
+func (a *agent) getDbInfo() string {
```
jakozaur, line 308: I'm debating whether we want to collect that. Maybe we should hash some of that, in particular URL may be sensitive in some env.

quesma/telemetry/phone_home.go
```diff
@@ -296,6 +296,63 @@ where active
...
+			return nil, fmt.Errorf("failed to scan row: %w", err)
+		}
+		tableSizes[tableName] = totalSize
+		allTablesSize += totalSize
+	}
+	tableSizes["QUESMA_all_tables_in_bytes"] = allTablesSize
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("error iterating over rows: %w", err)
+	}
+	return tableSizes, nil
```
jakozaur, line 353: We may have tons of tables and print them all; I'm not sure if this is desired. I would just print at most 100 of the biggest tables per size and total sum. I would also just print their size in MBs to prevent data leakage.

quesma/telemetry/phone_home.go
jakozaur: +100 Would prefer to have that in telemetry report.

jakozaur (commented):
Some suggestions.

mieciu:
@nablaone @jakozaur please take another look at this now, I moved all the data to telemetry's `stats` object

PR #1046, opened 2024-11-27
Adding quesma v2 core apis

pdelewski:
This PR adds APIs created in `quesma_api` POC to official codebase. According to our discussion I created separate v2 directory with new module


avelanarius (approved):
LGTM, `v2` directory seems like a good place

PR #1048, opened 2024-11-28
Add build info 

nablaone:


bin/build-image.sh
```diff
@@ -2,4 +2,9 @@
 # Rebuilds the Quesma Docker image and tags it as quesma:nightly
 set -e
 cd "$(dirname "$0/")/.."
-docker build -f quesma/Dockerfile -t quesma/quesma:nightly quesma
+
+QUESMA_BUILD_SHA=$(git rev-parse HEAD)
+QUESMA_BUILD_DATE=$(date -u +"%Y-%m-%d %H:%M:%S")
```
mieciu, line 7: I think that format should be aligned with what we set in official DockerHub releases, ref: https://github.com/QuesmaOrg/quesma/blob/main/.github/workflows/release.yml#L59

nablaone: Changed

mieciu (approved):

PR #1050, opened 2024-11-29
Fix incorrect field encodings for fields from mappings (but missing from JSON)

### LINKED ISSUE #1045: Incorrect field encoding for fields in ALTER TABLE
I don't yet have a reproducer for this problem, but I noticed that Quesma's ingest produced a following schema:

```
    `usergroup` Nullable(String) COMMENT 'quesmaMetadataV1:fieldName=usergroup',
```

However, the original data had a `userGroup` field, not `usergroup` field. It seems like the field encoding in the comment stored the name AFTER the encoding, not before the encoding.

Note that before ingestion I did a PUT mapping.

avelanarius:
In the `CREATE TABLE`, Quesma includes fields we got from `PUT /:index/_mapping` (but are missing from the JSON). This code path incorrectly computed field encodings - used the encoded field name in the comment instead of the field name before encoding.

Fix the issue by using the `reverseMap` (field name before encoding) in that code path.

Fixes #1045

### DIFF
--- a/ci/it/configs/quesma-ingest.yml.template
+++ b/ci/it/configs/quesma-ingest.yml.template
@@ -92,6 +92,9 @@ processors:
         nested_test:
           target:
             - my-clickhouse-instance
+        encodings_test:
+          target:
+            - my-clickhouse-instance
         "*":
           target:
             - my-minimal-elasticsearch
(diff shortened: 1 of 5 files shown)


pdelewski:
LGTM :)

pdelewski (approved):

PR #1054, opened 2024-11-29
Wait for HTTP 200 OK in integration tests

avelanarius:
We sometimes see integration tests failing, because Quesma returns HTTP 503 Service Unavailable and the test tries to use Quesma in such a unhealthy state. (The only place that 503 could originate from is the golang's http library code - in a case that a handler timeouts. I didn't think it was worthwhile to investigate it further, we only saw this issue in integration tests, not ever in a normal operation of Quesma).

This PR fixes the `WaitingFor` condition to wait for HTTP 200 OK for Elastic and Quesma containers. This should prevent the test from starting when the container is not yet fully ready.

### DIFF
--- a/ci/it/testcases/utils.go
+++ b/ci/it/testcases/utils.go
@@ -11,6 +11,7 @@ import (
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 	"log"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


pdelewski (approved):

avelanarius:
Actually now I see that `testcontainers` already did that check by default:

https://github.com/github/gh-ost/blob/067611679de3f445112112150c8614e0629c1a5d/vendor/github.com/testcontainers/testcontainers-go/wait/http.go#L65-L67

Therefore, this PR wouldn't actually help - Quesma in some cases first returns HTTP 200 OK, but then after a while it returns HTTP 503 Service Unavailable

So I'll have to investigate it further in the future...

PR #1065, opened 2024-12-05
PathRouter has to be moved up to v2

pdelewski:
In order to use it in v2 `quesma_api` I had to move it up to v2


mieciu (approved):
all good 🫡

PR #1067, opened 2024-12-05
Fix type of nested properties

jakozaur:
Fix two bugs:
- double quotes of some fields
- nested properties did not work with array or function, this fixes this problem


pdelewski:
LGTM :), unit test would be helpful

nablaone (approved):
LGTM

PR #1076, opened 2024-12-06
Fix "Field statistics" panel

nablaone:
This PR fixes the following cases:
1. simplify "exists" query parser
2. handle non-existing fields
3. handle falling back to the `attributes_values` column if available

2,3 may introduce a new kind of SQL error. Some functions do not handle `NULL` well.

There are skipped tests. @trzysiek  please take a look


trzysiek:
Unskipped all tests, there was a lot of them skipped :D
I'm almost positive we don't have any regression in tests because of the changes to them here, they might actually test a bit more (they are still *very* old cumbersome to modify, I won't lie).

trzysiek (approved):
It generally looks really nice, but it's too complicated for me to ensure it won't cause any regression. I'd need not-so-short amount of time to do some testing in Kibana dev console, which I don't have today.

My take is: we're positive this PR fixes a lot of views in Kibana, especially one which is important for the customer. I'd merge it today, and leave some more tests till I have time (tomorrow or Monday latest). If something breaks during that time, we'll know the cause, so we can easily revert.

I should be able to write some test cases then, out of my manual testing.

PR #1082, opened 2024-12-09
Fix Avg/Sum over Datetime column

nablaone:
Our queries fail if we try to do `sum` or `avg`  over a string or date column. This PR detects and compensates (adds a hack) it.

### DIFF
--- a/quesma/quesma/schema_transformer.go
+++ b/quesma/quesma/schema_transformer.go
@@ -724,6 +724,50 @@ func (s *SchemaCheckPass) convertQueryDateTimeFunctionToClickhouse(indexSchema s
 
 }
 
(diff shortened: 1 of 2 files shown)


trzysiek:
I think we should allow those functions on `DateTime*`, as they work in Elastic
Screenshot 2024-12-09 at 17 14 52
We even have support for that in code, e.g. in `model/metrics_aggregations/common.go` - the response json is different for dates than normally. It used to work, we must have some regression.

But anyway I'd do that in another PR. I can do that if you want, should take only a few mins.

trzysiek (approved):
Looks good! But if I understand correctly from the code, we will allow such a request, but will return `NULL` (result of `avg/sum(NULL)`)? 
It's fine, so far we probably can't do better, but I'll also change that soon to return error to Kibana just after parsing query (after https://github.com/QuesmaOrg/quesma/pull/1006 and some follow-up) so that we behave like Elastic

PR #1086, opened 2024-12-09
Fix override

jakozaur:
Overriding table name does not work, this fix it, but it was a hole of other issues. Got longer vision, but would love to validate before digging more in that direction.


quesma/clickhouse/clickhouse.go
```diff
@@ -177,6 +178,52 @@ func (lm *LogManager) ResolveIndexPattern(ctx context.Context, pattern string) (
 	return util.Distinct(results), nil
 }
 
+// We should use this instead of original code
+func (lm *LogManager) ResolveIndexPatternV2(schema schema.Registry, ctx context.Context, pattern string) (results []string, err error) {
```
nablaone, line 182: Why didn't you remove the previous version?

nablaone (approved):
LGTM

A single comment. I assume ResolveIndexPatternV2 is a temporary solution.

PR #1089, opened 2024-12-10
Add support for common table // kick out `TableResolver`

mieciu:
In this next-gen processor, we don't need an instance of table resolver. 

Each decision can be derived straight from the processor configuration.


pdelewski (approved):
Sounds good!

PR #1091, opened 2024-12-10
Add split_time_range optimizer

avelanarius:
Context: We noticed that for some schemas (that don't ORDER BY time), the "Discover" view in Kibana over long time ranges can be very slow, even though it only shows 500 results. Changing the time range to a shorter one can make the query faster. (See this issue in ClickHouse for a similar example: https://github.com/ClickHouse/ClickHouse/issues/69315)

This optimization therefore splits the time range into parts: a short time range, on which we bet that the query will be fast (and still return LIMIT many results) and a long time range, which will be used to get the rest of the results (in case the short time range didn't return enough results). The ranges are customizable (potentially even more than 2).

During the development I also fixed two related problems:
- added `AliasColumnsTransformation` which makes sure that the queries we generate have columns with clear names (so that I can reference them when they are in subquery)
- `read` from ClickHouse now only reads `LIMIT` many rows and closes `Rows` asynchronously (don't wait for query completion if we read `LIMIT` many rows)

### DIFF
--- a/quesma/clickhouse/quesma_communicator.go
+++ b/quesma/clickhouse/quesma_communicator.go
@@ -11,6 +11,7 @@ import (
 	"quesma/end_user_errors"
 	"quesma/logger"
 	"quesma/model"
+	"quesma/quesma/recovery"
 	tracing "quesma_v2/core/tracing"
 	"strconv"
 	"strings"
@@ -234,6 +235,7 @@ func read(ctx context.Context, rows *sql.Rows, selectFields []string, rowToScan 
 		return nil, fmt.Errorf("clickhouse: iterating over rows failed:  %v", rows.Err())
 	}
 	go func() {
+		recovery.LogPanicWithCtx(ctx)
 		err := rows.Close()
 		if err != nil {
 			logger.ErrorWithCtx(ctx).Msgf("clickhouse: closing rows failed: %v", err)
--- a/quesma/clickhouse/quesma_communicator.go
+++ b/quesma/clickhouse/quesma_communicator.go
@@ -59,10 +59,11 @@ func (lm *LogManager) ProcessQuery(ctx context.Context, table *Table, query *mod
 			colName = col.Alias
 		case model.LiteralExpr:
 
-			// This should be moved to the SchemaCheck pipeline. It'll require to change a lot of tests.
+			// There's now a AliasColumnsTransformation transformation that handles this,
+			// but it's not fully complete as it'll require to change a lot more tests.
 			//
-			// It can be removed just after the pancake will be the only way to generate SQL.
-			// Pancake SQL are aliased properly.
+			// The only remaining issue is that Pancake SQLs sometimes generate LiteralExpr in SELECT
+			// instead of ColumnRef (nested SQLs case).
 
 			if str, isStr := col.Value.(string); isStr {
 				if unquoted, err := strconv.Unquote(str); err == nil {
@@ -71,11 +72,15 @@ func (lm *LogManager) ProcessQuery(ctx context.Context, table *Table, query *mod
 					colName = str
 				}
 			} else {
+				// AliasColumnsTransformation should have handled this
+				logger.Warn().Msgf("Unexpected unaliased literal: %v", col.Value)
 				if colName == "" {
 					colName = fmt.Sprintf("column_%d", count)
 				}
 			}
 		default:
+			// AliasColumnsTransformation should have handled this
+			logger.Warn().Msgf("Unexpected unaliased literal: %v", col)
 			if colName == "" {
 				colName = fmt.Sprintf("column_%d", count)
 			}
@@ -188,7 +193,7 @@ func executeQuery(ctx context.Context, lm *LogManager, query *model.Query, field
 		performanceResult.Error = err
 		return nil, performanceResult, end_user_errors.GuessClickhouseErrorType(err).InternalDetails("clickhouse: query failed. err: %v, query: %v", err, queryAsString)
 	}
-	res, err = read(rows, fields, rowToScan)
+	res, err = read(ctx, rows, fields, rowToScan, query.SelectCommand.Limit)
 
 	elapsed := span.End(nil)
 	performanceResult.Duration = elapsed
@@ -204,7 +209,7 @@ func executeQuery(ctx context.Context, lm *LogManager, query *model.Query, field
 
 // 'selectFields' are all values that we return from the query, both columns and non-schema fields,
 // like e.g. count(), or toInt8(boolField)
-func read(rows *sql.Rows, selectFields []string, rowToScan []interface{}) ([]model.QueryResultRow, error) {
+func read(ctx context.Context, rows *sql.Rows, selectFields []string, rowToScan []interface{}, limit int) ([]model.QueryResultRow, error) {
 
 	// read selected fields from the metadata
 
@@ -213,7 +218,8 @@ func read(rows *sql.Rows, selectFields []string, rowToScan []interface{}) ([]mod
 		rowDb = append(rowDb, &rowToScan[i])
 	}
 	resultRows := make([]model.QueryResultRow, 0)
-	for rows.Next() {
+	// If a limit is set (limit != 0) then collect only the first 'limit' rows
+	for (len(resultRows) < limit || limit == 0) && rows.Next() {
 		err := rows.Scan(rowDb...)
 		if err != nil {
 			return nil, fmt.Errorf("clickhouse: scan failed: %v", err)
@@ -227,9 +233,11 @@ func read(rows *sql.Rows, selectFields []string, rowToScan []interface{}) ([]mod
 	if rows.Err() != nil {
 		return nil, fmt.Errorf("clickhouse: iterating over rows failed:  %v", rows.Err())
 	}
-	err := rows.Close()
-	if err != nil {
-		return nil, fmt.Errorf("clickhouse: closing rows failed: %v", err)
-	}
+	go func() {
+		err := rows.Close()
+		if err != nil {
+			logger.ErrorWithCtx(ctx).Msgf("clickhouse: closing rows failed: %v", err)
+		}
+	}()
 	return resultRows, nil
 }
--- a/quesma/optimize/split_time_range.go
+++ b/quesma/optimize/split_time_range.go
@@ -1,0 +1,287 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+package optimize
+
+import (
+	"fmt"
+	"quesma/logger"
+	"quesma/model"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Context: We noticed that for some schemas (that don't ORDER BY time), the "Discover" view in Kibana
+// over long time ranges can be very slow, even though it only shows 500 results. Changing the time range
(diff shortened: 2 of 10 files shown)


quesma/optimize/shorten_time_range.go
```diff
@@ -0,0 +1,248 @@
...
+		e.Left.Accept(b)
+		e.Right.Accept(b)
+		return e
+	}
+	selectCommand.Accept(visitor)
+
+	return lowerLimit, upperLimit
+}
+
+func (s shortenTimeRange) findTimeRange(selectCommand *model.SelectCommand) *timeRange {
```
jakozaur, line 87: Different names, check if possible to optimize and find time range.

quesma/optimize/shorten_time_range.go
```diff
@@ -0,0 +1,248 @@
...
+	if foundTimeRange == nil {
+		return query, nil
+	}
+
+	if foundTimeRange.direction != model.DescOrder {
+		// TODO: support other directions
+		return query, nil
+	}
+
+	splitPoint := foundTimeRange.upperLimit - 60*60*1000 // TODO: better split point
```
jakozaur, line 184: Move to constant and possible to override by configuration.

jakozaur (commented):
Let me try, looks like it might work.

quesma/quesma/search.go
```diff
@@ -254,9 +254,9 @@ func (q *QueryRunner) executePlan(ctx context.Context, plan *model.ExecutionPlan
 		return responseBody, err
 	}
 
-	if resp, err := q.checkProperties(ctx, plan, table, queryTranslator); err != nil {
-		return resp, err
-	}
+	//if resp, err := q.checkProperties(ctx, plan, table, queryTranslator); err != nil {
```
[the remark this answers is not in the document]
avelanarius: `checkProperties` has a comment that it's deprecated and should be removed. This had false positives in newly generated queries (`SELECT * FROM (... UNION ALL ...)`) since it tried to find columns from that subqueries in schema and failed (because they were aliased slightly differently).

quesma/optimize/shorten_time_range.go
avelanarius: Done

quesma/optimize/shorten_time_range.go
avelanarius: It's now configurable by YAML configuration:
```yaml
          optimizers:
            split_time_range:
              properties:
                ranges: "15,60,120"
```

quesma/optimize/split_time_range.go
```diff
@@ -0,0 +1,274 @@
...
+
+	logger.Debug().Msg("Query not eligible for time range optimization: ORDER BY not a column reference")
+	return "", model.DefaultOrder, false
+}
+
+func (s splitTimeRange) checkAndFindTimeLimits(selectCommand *model.SelectCommand, orderByColumnName string) (*timeRangeLimit, *timeRangeLimit) {
+	var lowerLimit, upperLimit *timeRangeLimit
+
+	visitor := model.NewBaseVisitor()
+	visitor.OverrideVisitInfix = func(b *model.BaseExprVisitor, e model.InfixExpr) interface{} {
```
jakozaur, line 80: Unfortunately, this does not seem to be correct. We should only check where part of the statement and top-level conditions are separated by AND. This does not support also case when there are multiple filtering of timestamp.

quesma/optimize/split_time_range.go
```diff
@@ -0,0 +1,274 @@
...
+			shorterTimeRangesMinutes = append(shorterTimeRangesMinutes, int64(parsedTimeRange))
+		}
+	}
+
+	result := []timeRangeLimit{foundTimeRange.lowerLimit}
+
+	for _, shorterTimeRangeMinute := range shorterTimeRangesMinutes {
+		var splitPoint timeRangeLimit
+		if foundTimeRange.upperLimit.funcName == "fromUnixTimestamp64Milli" {
+			splitPoint = timeRangeLimit{value: foundTimeRange.upperLimit.value - shorterTimeRangeMinute*60*1000, funcName: foundTimeRange.upperLimit.funcName}
```
jakozaur, line 173: Would use time constants.

quesma/optimize/split_time_range.go
```diff
@@ -0,0 +1,274 @@
...
+		} else {
+			splitPoint = timeRangeLimit{value: foundTimeRange.upperLimit.value - shorterTimeRangeMinute*60, funcName: foundTimeRange.upperLimit.funcName}
```
jakozaur, line 175: Would use time constants.

quesma/optimize/split_time_range.go
```diff
@@ -0,0 +1,274 @@
...
+	return result
+}
+
+func (s splitTimeRange) transformQuery(query *model.Query, properties map[string]string) (*model.Query, error) {
+	foundTimeRange := s.findTimeRange(&query.SelectCommand)
+	if foundTimeRange == nil {
+		return query, nil
+	}
+
+	if foundTimeRange.direction != model.DescOrder {
```
jakozaur, line 196: We should refactor those condition and put it in `findTimeRange`.

quesma/quesma/schema_transformer.go
```diff
@@ -845,6 +845,41 @@ func (s *SchemaCheckPass) checkAggOverUnsupportedType(indexSchema schema.Schema,
...
+			aliasedColumns[i] = aliasedExpr
+			continue
+		}
+		if _, ok := column.(model.FunctionExpr); ok {
+			aliasedColumns[i] = model.NewAliasedExpr(column, fmt.Sprintf("column_%d", i))
+			continue
+		}
+
+		aliasedColumns[i] = model.NewAliasedExpr(column, fmt.Sprintf("column_%d", i))
+		logger.Error().Msgf("Quesma internal error - unreachable code: unsupported column type %T", column)
```
jakozaur, line 873: I would really love context here, though not sure if in this PR.

jakozaur (commented):
Amazing work!

Several smaller, but one big item not mentioning, would be to write `split_time_range_test.go`. We really need to test that well.

quesma/optimize/split_time_range.go
```diff
@@ -0,0 +1,274 @@
...
+					upperBoundValue := functionExpr.Args[0].(model.LiteralExpr).Value.(int64)
+					upperLimit = &timeRangeLimit{value: upperBoundValue, funcName: functionExpr.Name}
+				}
+			case ">", ">=":
+				if functionExpr, ok := e.Right.(model.FunctionExpr); ok &&
+					(functionExpr.Name == "fromUnixTimestamp64Milli" || functionExpr.Name == "fromUnixTimestamp") {
+					lowerBoundValue := functionExpr.Args[0].(model.LiteralExpr).Value.(int64)
+					lowerLimit = &timeRangeLimit{value: lowerBoundValue, funcName: functionExpr.Name}
+				}
+			}
```
jakozaur, line 95: For example add here `case "AND":` and only then do `e.Left.Accept(b)`

jakozaur: Hmm `if e.Op = ...` later would be easier.

quesma/quesma/schema_transformer.go
```diff
@@ -845,6 +845,41 @@ func (s *SchemaCheckPass) checkAggOverUnsupportedType(indexSchema schema.Schema,
...
+			} else {
+				aliasedColumns[i] = col
+			}
+			continue
+		}
+		if aliasedExpr, ok := column.(model.AliasedExpr); ok {
+			aliasedColumns[i] = aliasedExpr
+			continue
+		}
+		if _, ok := column.(model.FunctionExpr); ok {
```
nablaone, line 867: This is tricky. If we have a query like "select column, count(*) from .."  the database will name the second column as 'count'. This code changes the name to 'column_1'.  Not sure how it affects the results rendering. 

Generally speaking, we should name every column on the parse step.

quesma/quesma/search.go
```diff
@@ -813,30 +792,6 @@ func (q *QueryRunner) Close() {
 	logger.Info().Msg("queryRunner Stopped")
 }
 
-func (q *QueryRunner) findNonexistingProperties(query *model.Query, table *clickhouse.Table, queryTranslator IQueryTranslator) []string {
```
nablaone, line 816: We should check if we are referring to the non-existing columns before executing the query.  Our customer monitors the query log table. It can be moved to the last step of the transformation pipeline.

quesma/quesma/schema_transformer.go
avelanarius: `the database will name the second column as 'count'. This code changes the name to 'column_1'.`

The existing code in `ProcessQuery` would represent unaliased `count` as `column_1` in `QueryResultRow`, so this naming is compatible with the existing code.

quesma/quesma/schema_transformer.go
```diff
@@ -845,6 +845,41 @@ func (s *SchemaCheckPass) checkAggOverUnsupportedType(indexSchema schema.Schema,
...
+		}
+
+		aliasedColumns[i] = model.NewAliasedExpr(column, fmt.Sprintf("column_%d", i))
+		logger.Error().Msgf("Quesma internal error - unreachable code: unsupported column type %T", column)
+	}
+	return aliasedColumns
+}
+
+func (s *SchemaCheckPass) applyAliasColumns(indexSchema schema.Schema, query *model.Query) (*model.Query, error) {
+	query.SelectCommand.Columns = columnsToAliasedColumns(query.SelectCommand.Columns)
```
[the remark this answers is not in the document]
avelanarius: Yes, it's desired - don't want to break CTEs and queries reading specific columns from them.

quesma/quesma/search.go
avelanarius: the logic would have to be extended for the case of CTEs and subqueries (it didn't work properly there):

```
SELECT columnNotInSchema FROM (SELECT columnInSchema AS columnNotInSchema FROM table)
```

Not sure if I should address it in this PR.

quesma/quesma/schema_transformer.go
avelanarius: Resolving as outside of scope of this PR.

quesma/clickhouse/quesma_communicator.go
```diff
@@ -227,9 +232,11 @@ func read(rows *sql.Rows, selectFields []string, rowToScan []interface{}) ([]mod
 	if rows.Err() != nil {
 		return nil, fmt.Errorf("clickhouse: iterating over rows failed:  %v", rows.Err())
 	}
-	err := rows.Close()
-	if err != nil {
-		return nil, fmt.Errorf("clickhouse: closing rows failed: %v", err)
-	}
+	go func() {
+		err := rows.Close()
```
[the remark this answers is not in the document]
avelanarius: Not sure what you mean by `panic`, added passing context in the amended commit.

avelanarius:
Addressed feedback comments (those that are resolved) in the amended commit.

quesma/clickhouse/quesma_communicator.go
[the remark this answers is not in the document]
jakozaur: If `rows.Close()` throws panic, whole application might crash.
```suggestion
		recovery.LogPanicWithCtx(ctx)
		err := rows.Close()
```

jakozaur (approved):
One tiny thing (add recovery for panic in go routine) and we are ready to merge 🚢 .

PR #1092, opened 2024-12-11
Creating 2 separate frontend connectors

pdelewski:


quesma/v2/core/mux.go
```diff
@@ -203,6 +203,14 @@ func (p *PathRouter) GetHandlers() map[string]HandlersPipe {
 }
 func (p *PathRouter) SetHandlers(handlers map[string]HandlersPipe) {
 	for path, handler := range handlers {
-		p.mappings = append(p.mappings, mapping{pattern: path, compiledPath: urlpath.New(path), handler: &handler})
+		if _, ok := handler.Predicate.(*predicateAlways); ok {
```
mieciu, line 206:
```suggestion
		if _, ok := handler.Predicate.(*predicateAlways); ok { // adding route in old API doesn't take predicates as parameter so we have to align handler registering like this
```

mieciu (approved):
👍

PR #1101, opened 2024-12-12
new aggregation IP Prefix

trzysiek:


quesma/model/bucket_aggregations/ip_prefix.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Current limitation: we expect Clickhouse field to be IPv4 (and not IPv6)
+
+// Clickhouse table to test SQLs:
+// CREATE TABLE __quesma_table_name (clientip IPv4) ENGINE=Log
+// INSERT INTO __quesma_table_name VALUES ('0.0.0.0'), ('5.5.5.5'), ('90.180.90.180'), ('128.200.0.8'),  ('192.168.1.67'), ('222.168.22.67')
+
+// TODO make part of QueryType interface and implement for all aggregations
+// TODO add bad requests to tests
+// Doing so will ensure we see 100% of what we're interested in in our logs (now we see ~95%)
```
jakozaur, line 21: FIY: Probably it make sense to log it once every few minutes. This could be very noisy.

quesma/model/bucket_aggregations/ip_prefix.go
```diff
@@ -0,0 +1,204 @@
...
+		return fmt.Errorf("params is not a map, but %+v", paramsRaw)
+	}
+
+	// check if required are present
+	for paramName, paramType := range requiredParams {
+		paramVal, exists := params[paramName]
+		if !exists {
+			return fmt.Errorf("required parameter %s not found in params", paramName)
+		}
+		if reflect.TypeOf(paramVal).Name() != paramType { // TODO I'll make a small rewrite to not use reflect here
```
jakozaur, line 46: Perhaps implement TODO?

quesma/model/bucket_aggregations/ip_prefix.go
```diff
@@ -0,0 +1,204 @@
...
+		"field":         "string",
+		"prefix_length": "float64", // TODO should be int, will be fixed
+	}
+	optionalParams := map[string]string{
+		"is_ipv6":              "bool",
+		"append_prefix_length": "bool",
+		"keyed":                "bool",
+		"min_doc_count":        "int",
+	}
+	logIfYouSeeThemParams := []string{"min_doc_count"}
```
jakozaur, line 33: In generalized version we could log unkown parameters too.

quesma/model/bucket_aggregations/ip_prefix.go
```diff
@@ -0,0 +1,204 @@
...
+		minDocCount:        minDocCount,
+	}
+}
+
+func (query *IpPrefix) AggregationType() model.AggregationType {
+	return model.BucketAggregation
+}
+
+func (query *IpPrefix) TranslateSqlResponseToJson(rows []model.QueryResultRow) model.JsonMap {
+	fmt.Println(rows)
```
jakozaur, line 101: Remove `fmt.Println`.

quesma/queryparser/pancake_aggregation_parser_buckets.go
```diff
@@ -359,6 +360,35 @@ func (cw *ClickhouseQueryTranslator) parseComposite(aggregation *pancakeAggregat
...
+
+	aggr := bucket_aggregations.NewIpPrefix(
+		cw.Ctx,
+		cw.parseFieldField(params, "ip_prefix"),
+		cw.parseIntField(params, "prefix_length", 0), // default doesn't matter, it's required
+		cw.parseBoolField(params, "is_ipv6", defaultIsIpv6),
+		cw.parseBoolField(params, "append_prefix_length", defaultAppendPrefixLength),
+		cw.parseBoolField(params, "keyed", defaultKeyed),
+		cw.parseIntField(params, "min_doc_count", defaultMinDocCount),
+	)
```
jakozaur, line 383: Looks nice, thinking about generic representation:
https://www.geeksforgeeks.org/reflect-tag-get-function-in-golang-with-examples/

quesma/model/bucket_aggregations/ip_prefix.go
trzysiek: unknown parameter in Kibana is always an error. We also error out here if we see it, here:
```
// check if only required/optional are present
for paramName := range params {
  if _, isRequired := requiredParams[paramName]; !isRequired {
    wantedType, isOptional := optionalParams[paramName]
    if !isOptional {
      return fmt.Errorf("unexpected parameter %s found in IP Range params %v", paramName, params)
    }
```
Or you meant something else @jakozaur ?

quesma/model/bucket_aggregations/ip_prefix.go
trzysiek: Let me do that in another PR, by this TODO I meant my generic type checker, which is big enough for a standalone PR. Even if your suggestion of using reflect is simpler, IMO it's still better to add that in another PR.

The code here fully works, so there shouldn't be a problem with waiting a bit for another PR.

PR #1102, opened 2024-12-12
Override cont

jakozaur:
Why:
- fixes broken `ci/it`
- actually maintaining in sync index name in two places proven to be harder
  - we are mixing reading configuration with runtime computed values, this got very problematic edge cases
  - cleaner solution is just to have it in one place

### DIFF
--- a/quesma/quesma/config/index_config.go
+++ b/quesma/quesma/config/index_config.go
@@ -21,7 +21,6 @@ type IndexConfiguration struct {
 	Target          any                               `koanf:"target"`
 
 	// Computed based on the overall configuration
-	Name         string
 	QueryTarget  []string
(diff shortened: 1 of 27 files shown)


quesma/quesma/config/index_config.go
```diff
@@ -31,21 +30,18 @@ type OptimizerConfiguration struct {
 	Properties map[string]string `koanf:"properties"`
 }
 
-func (c IndexConfiguration) TableName() string {
+func (c IndexConfiguration) TableName(origName string) string {
```
nablaone, line 33: What is an `origName` here? It looks suspicious.

nablaone (approved):

PR #1103, opened 2024-12-12
Integrate v1 reroute logic with v2 APIs

pdelewski:
This PR implements first step of integration, it calls `router.Matches` from request handler as in v1 version


mieciu (approved):
👍

Issue #1099, opened 2024-12-12
`search_after` parameter is not properly supported


avelanarius:
The "Discover" view in Kibana shows 500 results by default, but the user can request additional results by clicking "Load more":

<img width="1071" alt="image" src="https://github.com/user-attachments/assets/59278339-7ef7-49d8-a49e-bb88d6f9c08f" />

This makes Kibana send a query with `search_after` parameter in the query JSON body. Currently Quesma ignores this parameter, so the next pages don't load properly.

trzysiek:
Looking at it today

pdelewski:
Work in progress.

PR #1104, opened 2024-12-12
`search_after` param (used to paginate hits e.g. in `Discover`)

trzysiek:
https://github.com/QuesmaOrg/quesma/issues/1099

### DIFF
--- a/quesma/eql/query_translator.go
+++ b/quesma/eql/query_translator.go
@@ -105,7 +105,7 @@ func (cw *ClickhouseEQLQueryTranslator) parseQuery(queryAsMap types.JSON) (query
 	// no highlighting here
 	highlighter = queryparser.NewEmptyHighlighter()
 
-	searchQueryInfo.Typ = model.ListAllFields
+	searchQueryInfo.Type = model.ListAllFields
 
 	var eqlQuery string
 
--- a/quesma/model/query.go
+++ b/quesma/model/query.go
@@ -155,12 +155,13 @@ func (queryType HitsInfo) String() string {
 }
 
 type HitsCountInfo struct {
-	Typ             HitsInfo
+	Type            HitsInfo
 	RequestedFields []string
-	Size            int // how many hits to return
-	TrackTotalHits  int // >= 0: we want this nr of total hits, TrackTotalHitsTrue: it was "true", TrackTotalHitsFalse: it was "false", in the request
+	Size            int           // how many hits to return
+	TrackTotalHits  int           // >= 0: we want this nr of total hits, TrackTotalHitsTrue: it was "true", TrackTotalHitsFalse: it was "false", in the request
+	SearchAfter     []interface{} // used for pagination. Nil means it was empty in the request
 }
 
 func NewEmptyHitsCountInfo() HitsCountInfo {
-	return HitsCountInfo{Typ: Normal}
+	return HitsCountInfo{Type: Normal}
 }
--- a/quesma/queryparser/query_parser.go
+++ b/quesma/queryparser/query_parser.go
@@ -108,7 +108,7 @@ func (cw *ClickhouseQueryTranslator) ParseQuery(body types.JSON) (*model.Executi
 func (cw *ClickhouseQueryTranslator) buildListQueryIfNeeded(
 	simpleQuery *model.SimpleQuery, queryInfo model.HitsCountInfo, highlighter model.Highlighter) *model.Query {
 	var fullQuery *model.Query
-	switch queryInfo.Typ {
+	switch queryInfo.Type {
 	case model.ListByField:
 		// queryInfo = (ListByField, fieldName, 0, LIMIT)
 		fullQuery = cw.BuildNRowsQuery(queryInfo.RequestedFields, simpleQuery, queryInfo.Size)
@@ -1050,7 +1050,7 @@ func (cw *ClickhouseQueryTranslator) isItListRequest(queryMap QueryMap) (model.H
 
 	fields, ok := queryMap["fields"].([]any)
 	if !ok {
-		return model.HitsCountInfo{Typ: model.ListAllFields, RequestedFields: []string{"*"}, Size: size}, true
+		return model.HitsCountInfo{Type: model.ListAllFields, RequestedFields: []string{"*"}, Size: size}, true
 	}
 	if len(fields) > 1 {
 		fieldNames := make([]string, 0)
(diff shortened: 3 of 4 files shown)


quesma/model/search_after.go
```diff
@@ -0,0 +1,184 @@
...
+		Validate(searchAfter any) error
+		ApplyStrategyAndTransformQuery(query *Query, searchAfter any) *Query
+		DoSomethingWithHitsResult(hitsResult *SearchHits) // TODO change name
+	}
+	SearchAfterStrategyType int
+)
+
+func SearchAfterStrategyFactory(strategy SearchAfterStrategyType, timestampField ColumnRef) SearchAfterStrategy {
+	switch strategy {
+	case Foolproof:
```
jakozaur, line 26: Bulletproof seems a better name.

quesma/model/search_after.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return searchAfterParsedFoolproof{timestampMs: timestampMs, pkHashes: make([]string, 0)}, nil // TODO add parsing pk hashes
+}
+
+func (s SearchAfterStrategyFoolproof) ApplyStrategyAndTransformQuery(query *Query, searchAfterRaw any) *Query {
+	searchAfter, _ := s.validateAndParse(searchAfterRaw) // we validate during parsing and error there. No need to check here.
+	if searchAfter.timestampMs == emptySearchAfterTs {
+		return query
+	}
+	timestampRangeClause := NewInfixExpr(s.timestampField, "<=", NewFunction("fromUnixTimestamp64Milli", NewLiteral(searchAfter.timestampMs))) // TODO fix this for Hydrolix...
```
jakozaur, line 104: This is not always correct, as we might have timestamps in seconds, not milliseconds.

quesma/model/search_after.go
```diff
@@ -0,0 +1,184 @@
...
+	timestampField ColumnRef
+}
+
+func (s SearchAfterStrategyBasicAndFast) Validate(searchAfter any) error {
+	_, err := s.validateAndParse(searchAfter)
+	return err
+}
+
+func (s SearchAfterStrategyBasicAndFast) validateAndParse(searchAfter any) (timestampMs int64, err error) {
+	fmt.Println("searchAfter: ", searchAfter)
```
jakozaur, line 148: Remove.

quesma/model/search_after.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return emptySearchAfterTs, fmt.Errorf("for basic_and_fast strategy, search_after must be an integer")
+}
+
+func (s SearchAfterStrategyBasicAndFast) ApplyStrategyAndTransformQuery(query *Query, searchAfterRaw any) *Query {
+	searchAfterTs, _ := s.validateAndParse(searchAfterRaw) // we validate during parsing and error there. No need to check here.
+	if searchAfterTs == emptySearchAfterTs {
+		return query
+	}
+	timestampRangeClause := NewInfixExpr(s.timestampField, "<", NewFunction("fromUnixTimestamp64Milli", NewLiteral(searchAfterTs)))
```
jakozaur, line 175: Same here.

quesma/model/search_after.go
```diff
@@ -0,0 +1,184 @@
...
+func (s SearchAfterStrategyFoolproof) ApplyStrategyAndTransformQuery(query *Query, searchAfterRaw any) *Query {
+	searchAfter, _ := s.validateAndParse(searchAfterRaw) // we validate during parsing and error there. No need to check here.
+	if searchAfter.timestampMs == emptySearchAfterTs {
+		return query
+	}
+	timestampRangeClause := NewInfixExpr(s.timestampField, "<=", NewFunction("fromUnixTimestamp64Milli", NewLiteral(searchAfter.timestampMs))) // TODO fix this for Hydrolix...
+	logger.Debug().Msgf("search_after: %v, query before: %v", searchAfter, AsString(query.SelectCommand))
+	query.SelectCommand.WhereClause = And([]Expr{query.SelectCommand.WhereClause, timestampRangeClause})
+	query.SelectCommand.Limit += len(searchAfter.pkHashes)
+	logger.Debug().Msgf("query after: %v", AsString(query.SelectCommand))
```
jakozaur, line 108: There are two debugs, which, even if they are not rendered, they're still quite expensive SQL for every hit. Probably commenting them out would be the easiest solution.

jakozaur (commented):
Great work!

1. Having more generic timestamp method accomodating for different types would be awesome.
2. Not sure what if we don't have timestamp.
3. Tests.

quesma/model/search_after.go
trzysiek: It's always correct, because we read up the value from this:
```
func FormatSortValue(i any) any {
	switch v := i.(type) {
	case time.Time:
		// When returned as part of `sort`, timestamps are always returned as millis
		return v.UnixMilli()
	default:
		return i
	}
}
```
As you see, if it's present, it'll always be milliseconds

trzysiek: But I definitely agree that the way it works and why so is way too cryptic.

trzysiek:
> 2. Not sure what if we don't have timestamp.

Tested that, and
a) we will return empty `sort` field for every hit
b) even if we have timestamp, we can sort by some other field, and my solution won't work anyway

So unfortunately there's some TODO to do.

trzysiek:
> 1. Having more generic timestamp method accomodating for different types would be awesome.
Agreed, but let me do that in a quick next PR, this one already is quite big, wouldn't like to mess in a couple more places in this one.

quesma/quesma/schema_search_after_transformer.go
```diff
@@ -0,0 +1,198 @@
...
+	if !ok {
+		return nil, fmt.Errorf("search_after must be an array, got: %v", query.SearchAfter)
+	}
+
+	if len(asArray) != len(query.SelectCommand.OrderBy) {
+		return nil, fmt.Errorf("len(search_after) != len(sortFields), search_after: %v, sortFields: %v", asArray, query.SelectCommand.OrderBy)
+	}
+
+	return nil, nil
+	/*
```
jakozaur, line 74: Should we delete this comment or is it for bulletproof?

quesma/quesma/schema_search_after_transformer.go
```diff
@@ -0,0 +1,198 @@
...
+			return empty, fmt.Errorf("for Bulletproof strategy, search_after must be an integer")
+		}
+
+		return searchAfterParsedBulletproof{timestampMs: timestampMs, pkHashes: make([]string, 0)}, nil // TODO add parsing pk hashes
+
+	*/
+}
+
+func (s searchAfterStrategyBulletproof) transform(query *model.Query, searchAfterParsed []model.Expr) (*model.Query, error) {
+	//timestampRangeClause := NewInfixExpr(s.timestampField, "<=", NewFunction("fromUnixTimestamp64Milli", NewLiteral(searchAfter.timestampMs))) // TODO fix this for Hydrolix...
```
jakozaur, line 93: Same here.

jakozaur (approved):
I may need it to read one again, but good to merge. Thank you for your hard work.

PR #1105, opened 2024-12-13
Adding Helm chart for Quesma demo

mieciu:
Adding helm chart for Quesma demo on k8s.

More details in [README.md](https://github.com/QuesmaOrg/quesma/blob/aa05036c8497f0d648a20a1f22a71a56c59241b7/helm-chart/README.md)


helm-chart/quesma-demo/values.template.yaml
```diff
@@ -0,0 +1,171 @@
...
+        - name: elastic-query
+          type: elasticsearch-fe-query
+          config:
+            listenPort: 8080
+            disableAuth: true
+      backendConnectors:
+        - name: my-minimal-elasticsearch
+          type: elasticsearch
+          config:
+            url: #PLACE_YOUR_ELASTICSEARCH_URL_HERE
```
nablaone, line 138: I think this should point to the elasticsearch managed by this helm chart. Am I right?

helm-chart/quesma-demo/values.template.yaml
```diff
@@ -0,0 +1,171 @@
...
+        - name: my-sqldb-data-source
+          type: #PLACE_YOUR_CONNECTOR_TYPE_HERE - `clickhouse`, `hydrolix` or `clickhouse-os`
```
nablaone, line 140: Why there is no clickhouse in this helm chart?

helm-chart/quesma-demo/values.template.yaml
mieciu: correct! https://github.com/QuesmaOrg/quesma/pull/1105/commits/892eac8c71610067d03cda262d68b8307fb0bb21

nablaone (commented):
I've succesfuly installed it on the local K8S cluster (docker desktop)

Not sure about the purpose of this helm chart. If it is a demo we should add Clickhouse instance as well.

helm-chart/quesma-demo/values.template.yaml
mieciu: This chart was born as a demo for existing Hydrolix instances. 

ClickHouse being a stateful service is whole another ballgame when it comes to deploying it on k8s. I also have not been able to find any officially supported charts/operators, so I'm a little reluctant towards adding that.

nablaone (approved):
LGTM

PR #1108, opened 2024-12-13
Add scripts to run unit and integration tests.

jakozaur:
A simple short-hand instead of my bashrc. Others might find this useful.


mieciu (approved):
LGTM, although I personally like to kick them off from the IDE 😉

PR #1112, opened 2024-12-14
Support for IPv6 in IP Prefix aggr

trzysiek:
We had support for IPv4, but Jacek wanted for a) IPv6 and b) IPs stored in Clickhouse as ints.
This PR adds support for IPv6, and I think having IPs as ints in Clickhouse will also work, because whenever we query for IPs, we use `intDiv(ip_field, some_number)`. So we already make use of the fact that Clickhouse treats all IPs as ints (`IPv4` type: `uint32`, `IPv6` type: `uint128`)
Didn't test it yet, though.


quesma/clickhouse/util_test.go
trzysiek, line 1: Not sure if we want this in the repo, but I personally will find it very useful for toying with the Clickhouse driver.

jakozaur: I think it may make sense, I think also include it in the Quesma Admin Debug UI in long-run.

jakozaur (approved):
Love that.

Issue #1038, opened 2024-11-25
Not escaping `'` in `iLIKE` query properly


mieciu:
I noticed we don't escape `'` properly. 
I found this out today when messing with filters in `eCommerce Revenue Dashboard` in OpenSearch.

Screenshot telling more than 1000s of words:

<img width="2521" alt="image" src="https://github.com/user-attachments/assets/5f429702-034a-473c-8aa0-d0325e1934ab">

trzysiek:
Fixed with https://github.com/QuesmaOrg/quesma/pull/1114

PR #1114, opened 2024-12-15
Not escaping `'` in `iLIKE` query properly

trzysiek:
https://github.com/QuesmaOrg/quesma/issues/1038

As new tests show for a little bit, the issue seems fixed. Instead of doing nothing, I escape all `\` and `'` in all string literals. There might be some edge case where it's not the correct behaviour, but I thought for a bit and haven't found any.
Even in `regexp`, where we escape `_` (Clickhouse's equivalent of `.` in regex), escaping it twice because of this change (sending `\\_` to Clickhouse instead of `\_` before), seems to work just as fine.

Because of array joins issue, sample panel still doesn't work, but there are no more any escaping error messages, like in the issue description.
So, when running on `main`, we get a lot of:
```
quesma-1                 | Dec 22 15:36:13.876 ERR quesma/quesma/dual_write_proxy.go:252 > quesma request failed: Q3006: Database query has failed. You may get more details in the Quesma console.  clickhouse: query failed. err: code: 62, message: Syntax error: failed at position 488 ('(') (line 1, col 488): (((arrayJoin("category") __quesma_match 'Men's Clothing')
```
But when running on this branch, we only have errors like this (notice `'` properly escaped):
```
quesma-1                 | Dec 22 15:32:01.714 ERR quesma/quesma/dual_write_proxy.go:252 > quesma request failed: Q3006: Database query has failed. You may get more details in the Quesma console.  clickhouse: query failed. err: code: 62, message: Syntax error: failed at position 278 ('__quesma_match') (line 1, col 278): __quesma_match 'Men\'s Clothing')
```

**Additional info:**
We discussed with Rafał some other way to make it work: stop sending any strings in the query itself, but pass all of them as parameters, where the driver will take care of any escaping issues, etc.
It's probably a good idea for the future, but would require some non-trivial amount of work, unlike this straightforward solution, which I already had done.


quesma/quesma/search_test.go
```diff
@@ -299,6 +299,9 @@ func TestSearchHandler(t *testing.T) {
 
 	for i, tt := range testdata.TestsSearch {
 		t.Run(fmt.Sprintf("%s(%d)", tt.Name, i), func(t *testing.T) {
+			if i == 37 {
+				t.Skip("Regexp seems to be broken because of some transformations")
+			}
```
jakozaur, line 304: Shouldn't we fix this skip?

jakozaur (approved):
On potential comment.

quesma/quesma/search_test.go
trzysiek: It was actually my bad, it works.

PR #1117, opened 2024-12-16
Remove `clickhouse.LogManager` from `ClickhouseQueryTranslator`

mieciu:
This instance is not really needed, `HandleTermsEnum` was the only place where this has been used.

The biggest change is probably `terms_enum.go` and also the `ResolveField(cw.Ctx, fieldStr, cw.Schema)` - it's no longer a method, but function taking Schema as an argument.


trzysiek (approved):
Besides that 1 comment, as I don't think `ResolveField` almost exclusively based on schema, should be in `queryparser` package, cool, LGTM.

And thanks for giving me a hand in refactoring that `query_parser` stuff. It remembers oldest times of Quesma history, doesn't look great in comparison to most of our code 😆

PR #1124, opened 2024-12-18
v2 -  Named sub logger - logger refactoring

nablaone:
This PR adds the ability to inject a named sublogger into the component. The Logger interface is part of V2 API.

There are some TODOs, that will be done in the subsequent PRs.
- initialize logger
- global logger removal (if necessary)
- use component names from the config


quesma/v2/core/quemsa_logger.go
```diff
@@ -0,0 +1,146 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
```
pdelewski, line 1: Typo in file name, should be `quesma_logger.go` instead of `quemsa_logger.go`

pdelewski:
Made some tests. Looks good to me :)

pdelewski (approved):

quesma/v2/core/quemsa_logger.go
nablaone: Shame on me. Fixed.

PR #1127, opened 2024-12-18
Print container logs when integration test fails

avelanarius:
We see (relatively rarely) integration tests being flaky. In those cases of flakiness the current logging is not sufficient to really understand the cause of the problem.

This PR aims to improve logging by printing logs of all containers when a test fails.

`printContainerLogs` is added to `Containers.Cleanup()` code to print the logs of containers.

Additionally, the `Cleanup()` function is now **always** called, even if the containers failed to start or tests failed. To make sure it can print some logs:
- the code now has less panics, which would have prevented `Cleanup()` from running
- in `setupClickHouse()` and others we return the container object in case container failed to start instead of `nil` (but still with error) to be able to recover some logs
- if one container fails to start, `Containers` is now returned partially filled (but still with error), so that we can print some logs instead of giving up entirely

### DIFF
--- a/ci/it/testcases/utils.go
+++ b/ci/it/testcases/utils.go
@@ -55,7 +55,7 @@ func printContainerLogs(ctx context.Context, container *testcontainers.Container
 
 	log.Printf("Logs for container '%s':", name)
 	for _, line := range strings.Split(string(output), "\n") {
-		log.Printf("Container '%s': %s", name, line)
+		log.Printf("[%s]: %s", name, line)
 	}
 }
 
--- a/ci/it/testcases/utils.go
+++ b/ci/it/testcases/utils.go
@@ -10,10 +10,12 @@ import (
 	"github.com/docker/docker/api/types/container"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
 	"strings"
+	"testing"
 	"text/template"
 	"time"
 )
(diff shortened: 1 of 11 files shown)


ci/it/testcases/utils.go
```diff
@@ -34,19 +36,53 @@ type Containers struct {
...
+
+	output, err := io.ReadAll(reader)
+	if err != nil {
+		log.Printf("Failed to read logs for container '%s': %v", name, err)
+		return
+	}
+
+	log.Printf("Logs for container '%s':", name)
+	for _, line := range strings.Split(string(output), "\n") {
+		log.Printf("Container '%s': %s", name, line)
```
jakozaur, line 58: I would consider a more compact version, such as
```suggestion
		log.Printf("[%s]: %s", name, line)
```

ci/it/testcases/utils.go
```diff
@@ -34,19 +36,53 @@ type Containers struct {
...
+		return
+	}
+
+	output, err := io.ReadAll(reader)
+	if err != nil {
+		log.Printf("Failed to read logs for container '%s': %v", name, err)
+		return
+	}
+
+	log.Printf("Logs for container '%s':", name)
```
jakozaur, line 56: What if there are a lot of logs, wouldn't it make sense to limit it to certain amount of lines. E.g. last 100 or 1000. I've seen cases in past when logging broke everything.

ci/it/testcases/utils.go
```diff
@@ -34,19 +36,53 @@ type Containers struct {
...
-	if c.ClickHouse != nil {
-		(*c.ClickHouse).Terminate(ctx)
+}
+
+func (c *Containers) Cleanup(ctx context.Context, t *testing.T) {
+	if t.Failed() {
+		printContainerLogs(ctx, c.Elasticsearch, "Elasticsearch")
+		printContainerLogs(ctx, c.Quesma, "Quesma")
+		printContainerLogs(ctx, c.Kibana, "Kibana")
+		printContainerLogs(ctx, c.ClickHouse, "ClickHouse")
```
jakozaur, line 79: So far I only ever need Quesma logs, wouldn't it be too noisy? Maybe only print others if there is issue with container?

jakozaur (approved):
Love that, but got some comments.

ci/it/testcases/utils.go
avelanarius: We saw a couple IT fails where ClickHouse had problems. Similarly (maybe we haven't yet seen it in ITs) sometimes Elasticsearch has problems which causes problems with Quesma.

ci/it/testcases/utils.go
avelanarius: Improved the code to get the last 1MB of logs and 1000 lines from that buffer. This should be safe.

ci/it/testcases/utils.go
jakozaur: Ok.

PR #1129, opened 2024-12-19
Ingest - types testing 

nablaone:
This PR adds an ingest test to check what types are supported. The test is quite strict right now. It compares the ingested field value literally with the query result. A less strict human/Kibana-acceptable comparison will be added in the future.

It prints the following table:

Screenshot 2024-12-20 at 11 52 48


ci/it/configs/quesma-ingest.yml.template
```diff
@@ -97,7 +97,7 @@ processors:
             - my-clickhouse-instance
         "*":
           target:
-            - my-minimal-elasticsearch
+            - my-clickhouse-instance
```
pdelewski, line 100: Will this not change the default behavior?

ci/it/testcases/test_ingest.go
```diff
@@ -645,3 +654,423 @@ func (a *IngestTestcase) testFieldEncodingsMappingsBug(ctx context.Context, t *t
...
+
+		if r.claimedSupport && !r.currentSupport {
+			failedTypes = append(failedTypes, r.name)
+		}
+		if len(r.errors) > 0 {
+			fmt.Println("Type: ", r.name)
+			fmt.Println("Errors: ", strings.Join(r.errors, ", "))
+		}
+	}
+}
```
trzysiek, line 1076: We don't do anything with `failedTypes`. Maybe forgot to print?

ci/it/testcases/test_ingest.go
```diff
@@ -645,3 +654,423 @@ func (a *IngestTestcase) testFieldEncodingsMappingsBug(ctx context.Context, t *t
...
+					case string:
+						fieldValueAsString = strconv.Quote(v)
+					case float64:
+						fieldValueAsString = strconv.FormatFloat(v, 'f', -1, 64)
+
+					default:
+						data, err := json.Marshal(v)
+						if err != nil {
+
+						}
```
trzysiek, line 986: 😆

ci/it/testcases/test_ingest.go
```diff
@@ -645,3 +654,423 @@ func (a *IngestTestcase) testFieldEncodingsMappingsBug(ctx context.Context, t *t
...
+	if err != nil {
+		log.Fatalf("Error parsing JSON: %v", err)
+	}
+
+	// Extract and print the `fields` tree
+	for _, hit := range response.Hits.Hits {
+		return hit.Source
+	}
+	return nil
+}
```
trzysiek, line 684: Maybe move this code to some other file? Model, utils, or some other name?

When some (useful for other tests) classes are declared in the middle of a very long test file, it may be hard to find and reuse them

ci/it/testcases/test_ingest.go
```diff
@@ -645,3 +654,423 @@ func (a *IngestTestcase) testFieldEncodingsMappingsBug(ctx context.Context, t *t
...
+		Fields map[string][]string `json:"fields"`
+	}
+
+	type HitsWrapper struct {
+		Hits []Hit `json:"hits"`
+	}
+
+	type Response struct {
+		Hits HitsWrapper `json:"hits"`
+	}
```
trzysiek, line 699: Ah, and they are redeclared here ;) One of two declarations should probably go.

trzysiek:
I like this a lot! Left a few nitpicks myself + there's a more important Przemek's question.

ci/it/configs/quesma-ingest.yml.template
nablaone: I've moved this test to the separate test case. It has its own config now.

ci/it/testcases/test_ingest.go
nablaone: This is leftover. Test prints a summary table instead.

ci/it/testcases/test_ingest.go
nablaone: :D

trzysiek (approved):

PR #1135, opened 2024-12-21
Fix `ids` query DSL

trzysiek:
By accident I noticed our SQL generated for `ids` looks quite nonsensical, e.g. that's what's generated for my new very simple request:
```
WHERE "@timestamp" IN toDateTime64('2024-12-21 07:29:03.367', '2024-12-21 07:29:02.992', 3)
```
Also noticed we have 0 tests for this query. Fixing those 2 things here.
(adding `TupleExpr` for just this 1 usecase might seem unnecessary, but it's needed in my 2 other open PRs, so I'd do that)

Also, minor: changed `return model.NewSimpleQuery(nil, false)` to more descriptive `return model.NewSimpleQueryInvalid()` everywhere.

### DIFF
--- a/quesma/model/expr_string_renderer.go
+++ b/quesma/model/expr_string_renderer.go
@@ -4,6 +4,7 @@ package model
 
 import (
 	"fmt"
+	"quesma/logger"
 	"quesma/quesma/types"
 	"regexp"
 	"sort"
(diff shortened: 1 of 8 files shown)


quesma/model/expr_string_renderer.go
```diff
@@ -68,6 +69,22 @@ func (v *renderer) VisitLiteral(l LiteralExpr) interface{} {
 	return fmt.Sprintf("%v", l.Value)
 }
 
+func (v *renderer) VisitTuple(t TupleExpr) interface{} {
+	switch len(t.Exprs) {
+	case 0:
+		logger.WarnWithThrottling("VisitTuple", "TupleExpr with no expressions")
+		return "()"
```
pdelewski, line 76: Just to be consistent, consider returning `tuple()`

pdelewski:
LGTM :)

pdelewski (approved):

PR #1142, opened 2024-12-27
`search_after` param (used to paginate hits e.g. in Discover), 2/2 strategy 

trzysiek:


quesma/model/query_processor.go
```diff
@@ -37,26 +69,28 @@ func (qp *QueryProcessor) sameGroupByFields(row1, row2 QueryResultRow, level int
 				return false
 			}
 		} else if isArray1 && isArray2 {
-			return reflect.DeepEqual(val1, val2)
+			if !reflect.DeepEqual(val1, val2) {
+				return false
+			}
```
trzysiek, line 74: I think this was a bug.

quesma/model/query_processor.go
```diff
@@ -37,26 +69,28 @@ func (qp *QueryProcessor) sameGroupByFields(row1, row2 QueryResultRow, level int
...
 	}
 
 	lastRow := ResultSet[0]
 	buckets := [][]QueryResultRow{{lastRow}}
 	for _, row := range ResultSet[1:] {
-		if qp.sameGroupByFields(row, lastRow, level) {
+		if qp.sameFirstNColumns(row, lastRow, N) {
 			buckets[len(buckets)-1] = append(buckets[len(buckets)-1], row)
 		} else {
 			buckets = append(buckets, []QueryResultRow{row})
```
trzysiek, line 96: One of the last places where `level` name is used - getting rid of it.

PR #1146, opened 2024-12-30
Limit index  name length

nablaone:
ClickHouse stores column data in files, with each file’s name being a combination of the table name and the column name. We should not exceed the maximum path length. Elasticsearch limits the index name to 256 characters, and this pull request (PR) applies that limit.

Screenshot 2024-12-20 at 16 38 28

@pdelewski  please review this PR. Not sure how we should handle the field limit.

### DIFF
--- a/quesma/end_user_errors/clickhouse.go
+++ b/quesma/end_user_errors/clickhouse.go
@@ -48,6 +48,10 @@ func GuessClickhouseErrorType(err error) *EndUserError {
 			return ErrDatabaseTLSVerify.New(originalErr)
 		}
 
+		if strings.Contains(s, "code: 76") {
+			return ErrDatabaseStorageError.New(originalErr)
+		}
+
 		err = errors.Unwrap(err)
 		if err == nil {
 			break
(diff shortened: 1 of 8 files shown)


quesma/ingest/processor.go
```diff
@@ -688,6 +689,11 @@ func (ip *IngestProcessor) processInsertQuery(ctx context.Context,
 
 func (lm *IngestProcessor) Ingest(ctx context.Context, indexName string, jsonData []types.JSON) error {
 
+	err := elasticsearch.IsValidIndexName(indexName)
```
pdelewski, line 692: I'm wondering if we should check this limit here. A user could manually create an index with a name longer than 256 characters, and in that case, we would block ingestion into such indexes.

quesma/end_user_errors/clickhouse.go
```diff
@@ -48,6 +48,10 @@ func GuessClickhouseErrorType(err error) *EndUserError {
 			return ErrDatabaseTLSVerify.New(originalErr)
 		}
 
+		if strings.Contains(s, "code: 76") {
```
pdelewski, line 51: Maybe we could extract this `"code: 76"` as const

pdelewski:
Added minor comments, generally LGTM :)

pdelewski (approved):

quesma/ingest/processor.go
nablaone: Elastic blocks as well.

PR #1150, opened 2025-01-02
Fixing bind: address already in use

pdelewski:
This PR fixes problem mentioned in the title. It's not generic solution yet, however it good enough for now. I plan to work on generic solution in next PR

### DIFF
--- a/quesma/main.go
+++ b/quesma/main.go
@@ -209,7 +209,7 @@ func constructQuesma(cfg *config.QuesmaConfiguration, sl clickhouse.TableDiscove
 	if cfg.TransparentProxy {
 		return quesma.NewQuesmaTcpProxy(cfg, quesmaManagementConsole, logChan, false)
 	} else {
-		const quesma_v2 = true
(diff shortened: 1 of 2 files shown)


quesma/main.go
```diff
@@ -209,7 +209,7 @@ func constructQuesma(cfg *config.QuesmaConfiguration, sl clickhouse.TableDiscove
 	if cfg.TransparentProxy {
 		return quesma.NewQuesmaTcpProxy(cfg, quesmaManagementConsole, logChan, false)
 	} else {
-		const quesma_v2 = false
+		const quesma_v2 = true
```
mieciu, line 212: 💪

mieciu (approved):
totally fine for now! 👌

quesma/main.go
pdelewski: That's only to check if all tests are passing. It's not right time yet to do the switch, maybe next week. We should all agree :)

PR #1152, opened 2025-01-03
Use `-Array` variants of aggregates in `schema_array_transformer`

avelanarius:
`schema_array_transformer` transforms the SQL query for `Array` columns. Before this change, if an aggregation was performed on a `Array` column, e.g. `sum(myArrayColumn)`, the transformer would change it into `sum(arrayJoin(myArrayColumn))`.

However using `arrayJoin` function has problems - `arrayJoin` modifies the result set of SQL query introducing additional rows. If there are many `arrayJoin`s, a Cartesian product many rows will be performed: this causes query slowdown and makes the result invalid (we don't actually want to do a Cartesian product!).

Solve the problem by using `-Array` variants of aggregates (e.g. `sumArray` instead of `sum(arrayJoin())`), which does not inflate the number of result rows.

Note that this PR does NOT get rid of `arrayJoin()` fully in all cases. There are panels that actually need it, such as "Top products this week" in eCommerce dashboard, where we `GROUP BY` an array column. 

Screenshot 2025-01-07 at 11 20 42

This remaining case should use the `ARRAY JOIN` operator, but this is out-of-scope of this PR.

### DIFF
--- a/quesma/quesma/schema_array_transformer.go
+++ b/quesma/quesma/schema_array_transformer.go
@@ -16,6 +16,130 @@ import (
 //
 //
 
+// Aggregate functions names, generated from ClickHouse documentation:
+// git clone --depth 1 https://github.com/ClickHouse/ClickHouse.git
+// cd ClickHouse/docs/en/sql-reference/aggregate-functions/reference
+// find . -type f | cut -c3- | rev | cut -c4- | rev | sort
+
+var aggregateFunctions = map[string]bool{
+	"aggthrow":                     true,
+	"analysis_of_variance":         true,
+	"any":                          true,
+	"anyheavy":                     true,
+	"anylast":                      true,
+	"approxtopk":                   true,
+	"approxtopsum":                 true,
+	"argmax":                       true,
+	"argmin":                       true,
+	"arrayconcatagg":               true,
+	"avg":                          true,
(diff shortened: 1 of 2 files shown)


quesma/quesma/schema_array_transformer.go
```diff
@@ -81,13 +205,28 @@ func NewArrayTypeVisitor(resolver arrayTypeResolver) model.ExprVisitor {
 			if ok {
```
jakozaur, line 205: IF suffix ends with merge, do special logic.

quesma/quesma/schema_array_transformer.go
```diff
@@ -81,13 +205,28 @@ func NewArrayTypeVisitor(resolver arrayTypeResolver) model.ExprVisitor {
...
 				dbType := resolver.dbColumnType(column.ColumnName)
 				if strings.HasPrefix(dbType, "Array") {
-					if strings.HasPrefix(e.Name, "sum") {
```
nablaone, line 84: Maybe we should resurrect that idea 

https://github.com/QuesmaOrg/quesma/pull/666/files

jakozaur:
Closed in favor of:
https://github.com/QuesmaOrg/quesma/pull/1226

PR #1154, opened 2025-01-03
Enable async_insert optimizer by default

nablaone:
We should enable async_insert (#1151) by default. The default settings have been slightly changed to be more safe.

### DIFF
--- a/quesma/ingest/processor.go
+++ b/quesma/ingest/processor.go
@@ -785,9 +785,9 @@ func (ip *IngestProcessor) applyAsyncInsertOptimizer(tableName string, clickhous
 		// some sane defaults
 		clickhouseSettings["wait_for_async_insert"] = 1
 
-		clickhouseSettings["async_insert_busy_timeout_ms"] = 100
-		clickhouseSettings["async_insert_max_data_size"] = 100000
-		clickhouseSettings["async_insert_max_query_number"] = 1000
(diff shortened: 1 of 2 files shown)


quesma/ingest/processor.go
```diff
@@ -786,9 +784,10 @@ func (ip *IngestProcessor) applyAsyncInsertOptimizer(tableName string, clickhous
 
 		// some sane defaults
 		clickhouseSettings["wait_for_async_insert"] = 1
-		clickhouseSettings["async_insert_busy_timeout_ms"] = 1000
-		clickhouseSettings["async_insert_max_data_size"] = 1000000
-		clickhouseSettings["async_insert_max_query_number"] = 10000
+
+		clickhouseSettings["async_insert_busy_timeout_ms"] = 100
+		clickhouseSettings["async_insert_max_data_size"] = 100000
+		clickhouseSettings["async_insert_max_query_number"] = 1000
```
jakozaur, line 790: This looks quite small. 1MB is still small insert, so is 10k queries. Though I would keep 100ms so we don't add too much latency.

jakozaur (approved):

quesma/ingest/processor.go
nablaone: I've changed these limits above the default values.

PR #1119, opened 2024-12-17
Introduce Query Processor in v2 architecture

mieciu:

PR #1156, opened 2025-01-03
Get rid of additional QueryRunner

mieciu:
Changing the QueryRunner so that it relies on `clickhouse.LogManagerIFace` instead of `*clickhouse.LogManager`.

Having that, we won't need a duplicate of `QueryRunner` in https://github.com/QuesmaOrg/quesma/pull/1119. 

Caveat: we're adding a that dumb dummy table resolver, but that's way simpler than having a copy of QueryRunner. 

Lots of LoCs to save here 😉

### DIFF
--- a/quesma/table_resolver/table_resolver_dummy.go
+++ b/quesma/table_resolver/table_resolver_dummy.go
@@ -1,0 +1,39 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+
+package table_resolver
+
+import (
+	mux "quesma_v2/core"
+)
+
+// DummyTableResolver is a dummy implementation of TableResolver to satisfy the QueryRunner and make it be compatible with the v2 api
+// thanks to this we can reuse the existing QueryRunner implementation without any changes.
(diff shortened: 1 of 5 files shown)


quesma/table_resolver/table_resolver_dummy.go
```diff
@@ -0,0 +1,39 @@
...
+
+package table_resolver
+
+import (
+	mux "quesma_v2/core"
+)
+
+// DummyTableResolver is a dummy implementation of TableResolver to satisfy the QueryRunner and make it be compatible with the v2 api
+// thanks to this we can reuse the existing QueryRunner implementation without any changes.
+type DummyTableResolver struct{}
```
jakozaur, line 12: I would consider naming it `EmptyTableResolver` as it better describes and we already use that name.

jakozaur (approved):
Consider one rename.

quesma/table_resolver/table_resolver_dummy.go
mieciu: I'll handle this in the next PR as this will have to also undergo a change related to common table handling 👍

PR #1161, opened 2025-01-06
Gather functions on `DateTime(64)` to one place

trzysiek:
And deal with them in 1 place (a new transformation), instead of 10 different places and even modules.

I think this PR even fixes 2 small bugs, unknown for me beforehand.


platform/testdata/opensearch_requests.go
```diff
@@ -171,7 +171,7 @@ var OpensearchSearchTests = []SearchTestCase{
 			"track_total_hits": true
 		}`,
 		WantedSql: []string{
-			`("__timestamp">=fromUnixTimestamp64Milli(1712236698149) AND "__timestamp"<=fromUnixTimestamp64Milli(1712237598149))`,
+			`("-@timestamp">=fromUnixTimestamp64Milli(1712236698149) AND "-@timestamp"<=fromUnixTimestamp64Milli(1712237598149))`,
```
trzysiek, line 174: it's fine, WantedSql` = before transformation now, `WantedQueries` contains SQL with properly escaped `__timestamp`

platform/testdata/requests.go
trzysiek, line 1: Simplifying a few tests cases, pointless to test this `range: last 15min` filter or sth like that in every query.

pdelewski:
Smoke tests are failing. This PR does too many things at once. I think it would be good idea to split it to many smaller PRs and merge separately. For now, I'm going to close it.

PR #1163, opened 2025-01-07
Generalize `TableResolver` and `LogManager` implementation

mieciu:
... so that it relies on `quesma_api.BackendConnector` instead of `*sql.DB`.

This may look like it's complicating the current codebase, but effectively makes both of these structures reusable in the v2 API.

Ref: https://github.com/QuesmaOrg/quesma/pull/1119

### DIFF
--- a/quesma/telemetry/phone_home.go
+++ b/quesma/telemetry/phone_home.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"crypto/tls"
-	"database/sql"
 	"fmt"
 	"github.com/goccy/go-json"
 	"github.com/google/uuid"
@@ -20,6 +19,7 @@ import (
 	"quesma/elasticsearch"
 	"quesma/health"
 	telemetry_headers "quesma/telemetry/headers"
+	quesma_api "quesma_v2/core"
 	"quesma_v2/core/diag"
 	"sort"
 
@@ -64,7 +64,7 @@ type agent struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	clickHouseDb *sql.DB
+	clickHouseDb quesma_api.BackendConnector
 	config       *config.QuesmaConfiguration
 	clientId     string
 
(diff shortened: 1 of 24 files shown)


quesma/telemetry/phone_home.go
```diff
@@ -348,10 +348,10 @@ func (a *agent) CollectClickHouse(ctx context.Context) (stats diag.ClickHouseSta
 	// https://gist.github.com/sanchezzzhak/511fd140e8809857f8f1d84ddb937015
 	stats.Status = statusNotOk
 
-	dbStats := a.clickHouseDb.Stats()
+	//dbStats := a.clickHouseDb.Stats(). TODO: after moving to generic impl (quesma_api.BackendConnector) we lost this driver-specific call
```
pdelewski, line 351: I think, we should extend our API with `Stats` call

pdelewski:
Overall, LGTM, however smoke test is failing with `500`

quesma/telemetry/phone_home.go
mieciu: ++, though not sure yet about more general implementation - e.g. would it be fine to return just `map[string]any` or sth else

pdelewski: hmm, Maybe we could return something similar do `DBStats` https://pkg.go.dev/database/sql#DB.Stats

mieciu:
> Overall, LGTM, however smoke test is failing with `500`

Solved. 
Also adding IT seal of approval https://github.com/QuesmaOrg/quesma/actions/runs/12670871361

quesma/telemetry/phone_home.go
mieciu: https://github.com/QuesmaOrg/quesma/pull/1163/commits/64927185b408374183c6e765d0b0945692c3cfe3

mieciu:
https://github.com/QuesmaOrg/quesma/actions/runs/12672023503

pdelewski (approved):

PR #1164, opened 2025-01-08
Virtual table limit

nablaone:


pdelewski:
> We were limited to 1000 virtual tables. It's quite low if we want to store daily based indexes.

I see `100` not `1000` in the code. Is it an error in above comment or we talk about different limit?

pdelewski (approved):

PR #1167, opened 2025-01-09
Moving v2 (new quesma) tests

pdelewski:
This PR is about making some cleanup and moving v2 test to `quesma` package where, at the moment, is the right place as tests belongs to `quesma` as a library, not the main process.
It's also prerequisite to https://github.com/QuesmaOrg/quesma/pull/1166


mieciu (approved):
🚢

PR #1170, opened 2025-01-09
Multi search support 

nablaone:
This PR adds support for Multi Search API. It's quite rudimentary.  It's limited to querying clickhouse-based indexes only. 
The next PR will add support for querying all connectors and tests.


quesma/quesma/search.go
```diff
@@ -164,6 +166,106 @@ func (q *QueryRunner) HandleCount(ctx context.Context, indexPattern string) (int
...
+
+			if v, ok := line["index"].(string); ok {
+				currentQuery.indexName = v
+			} else {
+				currentQuery.indexName = defaultIndexName
+			}
+		} else {
+
+			q := types.JSON{}
+			q["query"] = line["query"]
```
pdelewski, line 193: Should not we checked whether `query` exists?

quesma/quesma/search.go
```diff
@@ -164,6 +166,106 @@ func (q *QueryRunner) HandleCount(ctx context.Context, indexPattern string) (int
...
+
+		if currentQuery == nil {
+			currentQuery = &msearchQuery{}
+
+			if v, ok := line["index"].(string); ok {
+				currentQuery.indexName = v
+			} else {
+				currentQuery.indexName = defaultIndexName
+			}
+		} else {
```
pdelewski, line 190: I think we could also add `continue` and remove `else` branch

quesma/quesma/search.go
nablaone: It's required. We should raise an error if not provided.

pdelewski: ok, but do we do that? I don't see it, maybe at different layer?

nablaone: I've just added it.

pdelewski (approved):

PR #1173, opened 2025-01-10
Omit non-existing indexes 

nablaone:
When we query multiple indexes, we may treat non-existent indexes as empty ones.  This solves the case when we store daily based indexes in the common table and query with Grafana.


mieciu (approved):
In general LGTM, what makes me a little worried is that neither of `search.go` related tests failed 🙃
I'm good with introducing one in a separate PR, so I'll leave the approval anyways, but could you please add one?

pdelewski:
> In general LGTM, what makes me a little worried is that neither of `search.go` related tests failed 🙃 I'm good with introducing one in a separate PR, so I'll leave the approval anyways, but could you please add one?

Agree, that we should invest some time in more tests.

PR #1175, opened 2025-01-10
Group indexes by prefix (common table)

nablaone:
When Grafana queries daily indexes:
- it adds a time range 
- it adds a list of the indexes within a time range 

So if we have a 3-year time range, Grafana will send 1k+ indexes in the query.  We transform that list into an `OR` expression, and hit the query limit. 

This PR adds an optimizer that transforms the  `OR` expressions into a single `starsWith` expression when indexes share the same prefix. 

If we have indexes with a pattern `daily-YYYYMMDD,` we can enable the optimizer: 
```
        '*':
          useCommonTable: true
          optimizers:
              - group_common_table_indexes:
                  disabled: false
                  properties:
                        "daily-": true
```


quesma/quesma/config/config_v2.go
```diff
@@ -727,6 +734,12 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 		conf.DefaultQueryTarget = defaultConfig.QueryTarget
 		conf.AutodiscoveryEnabled = slices.Contains(conf.DefaultQueryTarget, ClickhouseTarget)
 
+		if defaultIngestConfig, ok := queryProcessor.Config.IndexConfig[DefaultWildcardIndexName]; ok {
+			conf.DefaultQueryOptimizers = defaultIngestConfig.Optimizers
```
trzysiek, line 738: Did you forget to change name to `defaultQueryConfig`? If not, I don't know what's going on here 😆

nablaone: You're right. Fixed.

quesma/quesma/schema_transformer.go
```diff
@@ -356,6 +356,25 @@ func (s *SchemaCheckPass) applyMapTransformations(indexSchema schema.Schema, que
...
+func (s *SchemaCheckPass) computeListIndexPrefixesToGroup() []string {
+
+	const groupByCommonTableIndexes = "group_common_table_indexes"
+
+	var groupIndexesPrefix []string
+	if s.cfg.DefaultQueryOptimizers != nil {
+		if opt, ok := s.cfg.DefaultQueryOptimizers[groupByCommonTableIndexes]; ok {
+			if !opt.Disabled {
+				for k, v := range opt.Properties {
+					if v != "false" {
```
pdelewski, line 368: Can `v` be only `true` or `false`?

nablaone: It's a boolean flag for now.

quesma/quesma/schema_transformer.go
```diff
@@ -423,10 +442,40 @@ func (s *SchemaCheckPass) applyPhysicalFromExpression(currentSchema schema.Schem
...
 		if useCommonTable && from == physicalFromExpression {
 
-			var indexWhere []model.Expr
+			orExpression := make(map[string]model.Expr)
+
+			groupIndexesPrefix := s.computeListIndexPrefixesToGroup()
 
 			for _, indexName := range query.Indexes {
-				indexWhere = append(indexWhere, model.NewInfixExpr(model.NewColumnRef(common_table.IndexNameColumn), "=", model.NewLiteral(fmt.Sprintf("'%s'", indexName))))
+				var added bool
```
pdelewski, line 450: nitpick: This body could be separate function

pdelewski (approved):

PR #1179, opened 2025-01-13
Multiple schema resolve - speed up 

nablaone:
We can speed up the process of resolving multiple schemas more simply.  No cache is required now.

### DIFF
--- a/quesma/quesma/functionality/field_capabilities/field_caps.go
+++ b/quesma/quesma/functionality/field_capabilities/field_caps.go
@@ -49,7 +49,9 @@ func handleFieldCapsIndex(cfg map[string]config.IndexConfiguration, schemaRegist
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


quesma/quesma/functionality/field_capabilities/field_caps.go
```diff
@@ -49,7 +49,9 @@ func handleFieldCapsIndex(cfg map[string]config.IndexConfiguration, schemaRegist
 			continue
 		}
 
-		if schemaDefinition, found := schemaRegistry.FindSchema(schema.IndexName(resolvedIndex)); found {
+		schemas := schemaRegistry.AllSchemas()
```
pdelewski, line 52: Can we move it above, outside the loop?

nablaone: You're right.

PR #1187, opened 2025-01-14
WIP - Table discovery listeners 

nablaone:


quesma/schema/registry.go
```diff
@@ -74,6 +81,56 @@ func (s *schemaRegistry) getInternalToPublicFieldEncodings(tableName string) map
...
+
+	protectedReload := func() {
+		defer recovery.LogPanic()
+		s.Lock()
+		defer s.Unlock()
+
+		s.invalidateCache()
+	}
+
+	go func() {
```
pdelewski, line 102: This starts a new goroutine. I think we should also provide a stop method for graceful shutdown.

quesma/schema/registry.go
```diff
@@ -74,6 +81,56 @@ func (s *schemaRegistry) getInternalToPublicFieldEncodings(tableName string) map
 	return internalToPublicFieldsEncodings
 }
 
+func (s *schemaRegistry) invalidateCache() {
+	s.cachedSchemas = nil
+}
+
+func (s *schemaRegistry) start() {
```
pdelewski, line 88: For consideration, should not this be public method, explicitly invoked in main() similarly to other components?

quesma/schema/registry.go
```diff
@@ -126,46 +183,57 @@ func (s *schemaRegistry) populateSchemaFromDynamicConfiguration(indexName string
...
-		logger.Error().Msgf("error loading schemas: %v", err)
-		return Schema{}, false
-	} else {
-		schema, found := schemas[name]
-		return schema, found
-	}
+	s.Lock()
+	defer s.Unlock()
+
+	schemas := s.loadOrGetSchemas()
```
pdelewski, line 196: `loadOrGetSchemas()` might return nil. I think we should check against this condition

quesma/schema/registry.go
```diff
@@ -74,6 +81,56 @@ func (s *schemaRegistry) getInternalToPublicFieldEncodings(tableName string) map
...
+
+}
+
+func (s *schemaRegistry) loadOrGetSchemas() map[IndexName]Schema {
+
+	if s.cachedSchemas == nil {
+		schema, err := s.loadSchemas()
+		if err != nil {
+			logger.Error().Err(err).Msg("error loading schema")
+			return nil
```
pdelewski, line 126: Maybe, we should return empty `map`

quesma/schema/registry.go
nablaone: It follows the Start/Stop pattern now.

pdelewski (approved):

PR #1192, opened 2025-01-16
Router fixes

mieciu:
There are two big changes here: 
* the algorithm which merges routes when there are multiple fronted connectors (individual routes used to overwrite themselves)
* `findHandler` in `quesma/v2/core/mux.go` changes, because it had pretty nasty issue that for given two mappings with the same paths it just took the first one, and if predicate didn't match - request was classified as non matching.

**WARNING:** This changes the behavior of the **currently used router**.


quesma/v2/core/quesma_builder.go
```diff
@@ -145,7 +142,7 @@ func (quesma *Quesma) buildInternal() (QuesmaBuilder, error) {
 						if info.pipelineIndex == pipelineIndex && info.connIndex == connIndex {
 							continue
 						}
-						quesma.dependencies.Logger().Info().Msgf("Sharing frontend connector %p with %p", sharedFc, pipeline.GetFrontendConnectors()[connIndex])
+						quesma.dependencies.Logger().Info().Msgf("Sharing frontend connector %v with %v", sharedFc.InstanceName(), conn.InstanceName())
```
pdelewski, line 145: I would prefer to leave old version

pdelewski (approved):

mieciu:
ITs: https://github.com/QuesmaOrg/quesma/actions/runs/12830773174

pdelewski:
LGTM :)

pdelewski (approved):

PR #1202, opened 2025-01-20
Necessary fixes for `_bulk` endpoint

mieciu:
For our V2 ingest processor, we need better implementation of `_bulk` API, which can handle more situations. Some shortcuts we took earlier, are no longer a viable option.

Changes:
* We used to take bulk payload lines and just process them in pairs (action/metadata line followed by document) which doesn't have to be the case. There can always be a `delete` call which isn't followed by any document. So the whole `BulkForEach` had to be rewritten.
* Extending `NDJSON` type validation, so that single-line JSON is also a valid NDJSON 😉 
* Adding some tests 

When loading sample datasets and having operational Kibana, every single request ends with code 200 🎊


quesma/quesma/functionality/bulk/bulk.go
```diff
@@ -104,16 +104,23 @@ func Write(ctx context.Context, defaultIndex *string, bulk types.NDJSON, ip *ing
...
 
-	return results, nil
+	// Here we filter out empty results so that final response does not contain empty elements
+	nonEmptyResults := make([]BulkItem, 0, len(results))
+	for _, result := range results {
+		if result != (BulkItem{}) {
+			nonEmptyResults = append(nonEmptyResults, result)
+		}
+	}
+	return nonEmptyResults, nil
```
avelanarius, line 117: alternatively `SplitBulk` could return properly-sized `results`

avelanarius (approved):

quesma/quesma/functionality/bulk/bulk.go
mieciu: Yea, as much as I don't like the current approach with shared array flying around between methods, I purposefully  avoided changing that in this PR because of the amount of changes I had to make so far.

I'll place a little TODO though.

mieciu:
IT run https://github.com/QuesmaOrg/quesma/actions/runs/12871349933

PR #1204, opened 2025-01-21
v2 extending base processor api

pdelewski:
This PR is about extracting/using common processing algorithm inside `baseprocessor`. This is yet another step forward and on-going effort.


quesma/model/transformation_pipeline.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func NewTransformationPipeline() *TransformationPipeline {
+	return &TransformationPipeline{}
+}
+
+func (o *TransformationPipeline) Transform(queries []*Query) ([]*Query, error) {
+	var err error
+	for _, transformer := range o.transformers {
+		queries, err = transformer.Transform(queries)
```
mieciu, line 16: perhaps we could leverage `multierror` in this case instead of just returning first one?

mieciu (commented):
One suggestion about using multierror for transformation pipeline, other than that -LGTM 💪

mieciu (approved):
(procedurally unblocking)

quesma/model/transformation_pipeline.go
pdelewski: fixed

PR #1211, opened 2025-01-22
Add (TCP) passthrough MySQL processor

avelanarius:
This PR adds a new processor and frontend/backend connectors that allows Quesma to act as a MySQL proxy.

Added components:
- `TcpMysqlConnectionHandler` - splits incoming TCP stream into MySQL wire protocol packets
- `TcpPassthroughMysqlProcessor` - reads incoming MySQL packets, forwards them to a real MySQL server (via `TcpBackendConnector`) and forwards back any responses from the real MySQL server to the connected client.
- `TcpBackendConnector` - a raw TCP connection to some server (in this case MySQL server)

This is a very early experimental version. In particular it has (at least) these limitations:
- No support for encrypted MySQL traffic
- Manually parses MySQL packets - we should probably use Vitess for that
- Stores the connection inside metadata, not some `Session` abstraction
- Doesn't support asynchronous packets from MySQL server - we read from the real MySQL server only after receiving some client request
- Reading from real MySQL server is very naive - it tries to read as much as possible in a loop with 1s deadline. When 1s deadline passes, it assumes that MySQL server sent everything
- TCP connection to real MySQL server is not closed properly (leak)

You can try the experiment by setting `mysql_passthrough_experiment` to `true` in `main.go`. It assumes that MySQL is running at `localhost:3306` and it will expose Quesma at `localhost:13306`. You can connect to it with `mysql` CLI tool:
```
docker run --name some-mysql -e MYSQL_ROOT_PASSWORD=my-secret-pw -p 3306:3306 -d mysql:latest

# Start Quesma (mysql_passthrough_experiment = true in code)

mysql --skip-ssl -h localhost -P 13306 -u root -p
```

### DIFF
--- a/quesma/frontend_connectors/tcp_mysql_connection_handler.go
+++ b/quesma/frontend_connectors/tcp_mysql_connection_handler.go
@@ -1,5 +1,6 @@
 // Copyright Quesma, licensed under the Elastic License 2.0.
 // SPDX-License-Identifier: Elastic-2.0
+// Experimental alpha frontend for MySQL protocol
 
 package frontend_connectors
 
--- a/quesma/frontend_connectors/tcp_mysql_connection_handler.go
+++ b/quesma/frontend_connectors/tcp_mysql_connection_handler.go
@@ -1,0 +1,107 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+
+package frontend_connectors
+
+import (
+	"encoding/binary"
+	"fmt"
+	"github.com/QuesmaOrg/quesma/quesma/backend_connectors"
+	quesma_api "github.com/QuesmaOrg/quesma/quesma/v2/core"
+	"io"
+	"net"
+)
+
+type TcpMysqlConnectionHandler struct {
+	processors []quesma_api.Processor
+}
+
+var ErrInvalidPacket = fmt.Errorf("invalid packet")
+
+func ReadMysqlPacket(conn net.Conn) ([]byte, error) {
+	// MySQL wire protocol packet format (see https://dev.mysql.com/doc/dev/mysql-server/8.4.3/PAGE_PROTOCOL.html):
+	// - 3 bytes: length of the packet (= LEN)
+	// - 1 byte: sequence ID
+	// - LEN bytes: packet body
+	//
+	// TODO: when packet is larger than 16MB, it's split into multiple packets. This code does NOT support this case yet.
+
+	packetLengthBytes, err := backend_connectors.ConnRead(conn, 3)
+	if err == io.EOF {
+		return nil, err
+	}
+	if err != nil || len(packetLengthBytes) != 3 {
+		return nil, ErrInvalidPacket
+	}
+	packetLength := int(binary.LittleEndian.Uint32(append(packetLengthBytes, 0)))
+
+	sequenceId, err := backend_connectors.ConnRead(conn, 1)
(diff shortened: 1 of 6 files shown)


pdelewski (approved):

quesma/frontend_connectors/tcp_mysql_connection_handler.go
```diff
@@ -0,0 +1,107 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+
```
jakozaur, line 3: Maybe we should add a comment.
```suggestion
// Experimental alpha frontend for Mysql protocol

```

quesma/frontend_connectors/tcp_mysql_connection_handler.go
```diff
@@ -0,0 +1,107 @@
...
+		var message any
+
+		metadata, message = dispatcher.Dispatch(p.processors, metadata, nil)
+		if message != nil {
+			_, err := conn.Write(message.([]byte))
+			if err != nil {
+				return fmt.Errorf("error sending response: %w", err)
+			}
+		}
+	}
```
jakozaur, line 78: It's not a Golang guru, but is it a common pattern to have curly bracket like that? Maybe we just inline here; then, we also avoid the need of redeclaring the message later.

quesma/main.go
```diff
@@ -53,6 +57,12 @@ const EnableConcurrencyProfiling = false
 //}
 
 func main() {
+	const mysql_passthrough_experiment = false
```
jakozaur, line 60:
```suggestion
        // TODO: Experimental feature, move to the configuration after architecture v2
	const mysql_passthrough_experiment = false
```

quesma/processors/tcp_passthrough_mysql_processor.go
```diff
@@ -0,0 +1,115 @@
+// Copyright Quesma, licensed under the Elastic License 2.0.
+// SPDX-License-Identifier: Elastic-2.0
+
```
jakozaur, line 3:
```suggestion
// Experimental alpha processor for Mysql protocol

```

jakozaur (approved):
Great work!

quesma/frontend_connectors/tcp_mysql_connection_handler.go
avelanarius: I'll leave it as-is for now. In the future we should remove sending that dummy `nil` (that curly bracket block) altogether, so I didn't bother factoring it out at this stage.

avelanarius:
> - How can we incorporate ConnWrite(conn net.Conn, data []byte) error and ConnRead(conn net.Conn, data []byte) error into the BackendConnector interface?

`ConnWrite` and `ConnRead` are relevant only to the TCP protocol, so I don't think they should be a part of `BackendConnector` interface.

`ConnWrite` and `ConnRead` are only helper functions to operate of `net.Conn` and the connector user in my opinion would prefer to operate on `net.Conn` itself, not our wrappers around it. The only reason for writing our own "wrappers" would be to have a unified interface around different connectors, but `net.Conn` and `sql.DB` are already "unified enough" (e.g. `sql.DB` supporting many SQL DBs).

> - Can we somehow standardize the HandleConnection flow to be common for all SQL connection handlers?
> - Similarly, can we standardize Processor::Handle in the context of common connection handling?

That's a good idea, but something to research in the (near) future.

PR #1212, opened 2025-01-22
Initial version of MySQL frontend connector (based on Vitess)

avelanarius:
In #1211, an initial (experimental) version of (TCP) passthrough MySQL processor was introduced. That processor/frontend was merely a minimally invasive TCP proxy, which snooped on the traffic, was able to recognize Query packets and potentially modify some packets. That processor/frontend required us to always have a real MySQL instance running and it was not easy for the processor to use a different SQL DB as a backend (such as Postgres).

The components in this PR are MySQL frontend connector (based on Vitess) and MySQL processor (paired with MySQL Vitess frontend connector). They try to emulate a MySQL instance and don't require having a real MySQL instance running. A disadvantage of this approach is that we have to manually construct MySQL responses (abstracted by Vitess structures).

This is still an experimental frontend/processor (alpha) and it's only a base code, on which we will further experiment.


quesma/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/QuesmaOrg/quesma/quesma
 
-go 1.23.2
+go 1.23.5
```
avelanarius, line 3: Vitess library requires go 1.23.5

quesma/frontend_connectors/vitess_mysql_connector.go
```diff
@@ -0,0 +1,166 @@
...
+		defer recovery.LogPanic()
+		t.listener.Accept()
+	}()
+	return nil
+}
+
+// Implementation of Vitess mysql.Handler interface:
+
+type ComQueryMessage struct {
+	Conn  *mysql.Conn
```
pdelewski, line 58: Seems unused for now.  This makes `frontend_connector` and `processor` strongly coupled. I would suggest to remove it for now

avelanarius: Removed it in https://github.com/QuesmaOrg/quesma/pull/1212/commits/4d23a21ec3811df6e283e717e4f67b633213f446

pdelewski (approved):

Issue #1199, opened 2025-01-20
@timestamp DateTime64(9)


kota-shade:
It's a question.
timestamp DateTime64(3) is not enough for id in my DB.
Sometimes there are many rows with the same timestamp.

But when I create table with @timestamp DateTime64(9) kibana doesn't show JSON tab.
The query WHERE is: 
WHERE ("@timestamp" = toDateTime64('2025-01-16 12:18:15.735013156',3) AND 0=0

And as the result the value is rounding up 3 digits after point and record is not found.
I don't khow Go. Is '3' hardcoded?
I can find something like this in source:
 return model.NewFunction("toDateTime64", model.NewLiteral(id), model.NewLiteral(3))

What is the solution for me?

trzysiek:
Hi @kota-shade !
Thanks a lot for the issue and I'm sorry that we indeed had a bug. I'm 99% sure it's fixed now after https://github.com/QuesmaOrg/quesma/pull/1214 . I'll confirm later today after I'll reproduce your case manually.

PR #1214, opened 2025-01-23
Better support for `DateTime(X)`, where `X != 3`

trzysiek:
Fix for https://github.com/QuesmaOrg/quesma/issues/1199

Now it seems to work with Clickhouse well. I'll do a full short manual test in a second, but I'm 99% the issue is fixed.


quesma/util/dates_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+func TestFindTimestampPrecision(t *testing.T) {
+	var testcases = []struct {
+		timestamp         string
+		expectedPrecision int
+		expectedSuccess   bool
+	}{
+		{"2024-12-21 07:29:03.3", 1, true},
+		{"2024-12-21 07:29:03", 0, true},
+		{"2024-12-21 07:29:3.3312", -1, false}, // we expect :03, not :3
```
pdelewski, line 18: Don't understand this case

pdelewski:
LGTM :)

quesma/util/dates.go
```diff
@@ -2,10 +2,56 @@
...
+			return -1, false
+		} else {
+			for i := lastDot + 1; i < len(timestamp); i++ {
+				if timestamp[i] < '0' || timestamp[i] > '9' {
+					return -1, false
+				}
+				precision += 1
+			}
+		}
+		return precision, true
```
jakozaur, line 55: We may consider returning: `len(timestamp)-lastDot` and don't bother adding precision.

jakozaur (approved):

PR #1223, opened 2025-01-23
Bring back ingest ab result to CH

mieciu:
I decided to revert this change and make it work in V2 architecture. 

Little nonsense in the dependencies, but at the cost of 100% feature parity 🙃


mieciu:
IT run https://github.com/QuesmaOrg/quesma/actions/runs/12930719421

pdelewski (approved):

PR #1224, opened 2025-01-24
Old admin console in v2 api

mieciu:
Another round of v2 efforts. 

This time the only "invasive" part is commenting out one validation which is incompatible with v2 realm.


quesma/quesma/config/config_v2.go
```diff
@@ -614,11 +695,12 @@ func (c *QuesmaNewConfiguration) TranslateToLegacyConfig() QuesmaConfiguration {
 					}
 
 				}
-				if len(processedConfig.QueryTarget) == 2 && !((processedConfig.QueryTarget[0] == ClickhouseTarget && processedConfig.QueryTarget[1] == ElasticsearchTarget) ||
-					(processedConfig.QueryTarget[0] == ElasticsearchTarget && processedConfig.QueryTarget[1] == ClickhouseTarget)) {
-					errAcc = multierror.Append(errAcc, fmt.Errorf("index %s has invalid dual query target configuration", indexName))
-					continue
-				}
+				// TODO: can we live without this?
```
mieciu, line 669: FYI @avelanarius ☝️

quesma/backend_connectors/clickhouse_backend_connector.go
```diff
@@ -4,23 +4,27 @@
...
+	"github.com/QuesmaOrg/quesma/quesma/buildinfo"
+	"github.com/QuesmaOrg/quesma/quesma/quesma/config"
 
 	quesma_api "github.com/QuesmaOrg/quesma/quesma/v2/core"
 )
 
 type ClickHouseBackendConnector struct {
 	BasicSqlBackendConnector
-	Endpoint string
+	//Endpoint string
```
pdelewski, line 18: leftover?

mieciu: indeed!

quesma/backend_connectors/elasticsearch_backend_connector.go
```diff
@@ -41,6 +42,24 @@ func NewElasticsearchBackendConnector(cfg config.ElasticsearchConfiguration) *El
 	return conn
 }
 
+// NewElasticsearchBackendConnector2 is an alternative constructor which uses the generic database configuration object
+func NewElasticsearchBackendConnector2(cfg config.RelationalDbConfiguration) *ElasticsearchBackendConnector {
```
pdelewski, line 46: Maybe 
```suggestion
func NewElasticsearchBackendConnectorWithDbConfig(cfg config.RelationalDbConfiguration) *ElasticsearchBackendConnector {
```

pdelewski, line 46: I'm wondering if we really need 2 constructors. Maybe we can create 
```
config: config.ElasticsearchConfiguration{
			Url:      cfg.Url,
			User:     cfg.User,
			Password: cfg.Password,
		},
```
outside and pass it to first constructor?

pdelewski:
Some minors, overall LGTM :)

pdelewski (approved):

quesma/backend_connectors/elasticsearch_backend_connector.go
mieciu: The separation comes from the fact that I'd like to eventually get rid of `config.ElasticsearchConfiguration` (and ideally all of the legacy config. So I thought it could be nicer in the long run to just kick out the second constructor with all its usages further down the road

PR #1227, opened 2025-01-24
Clear `CLIENT_SSL` in MySQL `Handshake` packet

avelanarius:
The `TcpPassthroughMysqlProcessor` does not yet support SSL, so when using a MySQL client you had to explicitly disable SSL on the client side or disable it on the MySQL server.

This change rewrites the initial `Handshake` packet sent by the MySQL server to inform the client that the server doesn't support SSL. After this change you can connect to MySQL via Quesma without having to manually disable SSL on the client side:

```
mysql -h localhost -P 13306 -u root -p
```

instead of

```
mysql --skip-ssl -h localhost -P 13306 -u root -p
```

### DIFF
--- a/quesma/processors/tcp_passthrough_mysql_processor.go
+++ b/quesma/processors/tcp_passthrough_mysql_processor.go
@@ -5,10 +5,12 @@
 package processors
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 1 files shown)


avelanarius:
It's now a part of https://github.com/QuesmaOrg/quesma/pull/1227, so closing this.

PR #1230, opened 2025-01-26
Turing-4 no `aggregation` in response

trzysiek:
Such a simple query
```
/async_search
"aggs": {
        "2": {
            "terms": {
                "field": "path_id",
                "order": {
                    "_count": "desc"
                },
                "size": 20
            }
        }
    },
    "query": {
       ..
    }
}
```
Gives no 'aggregation' part of response. It was a one-line fix in `terms`.
 
---------------------------------------
Old description, seems irrelevant to me ATM, let's come back to it once it's a problem.
 
```
"hits": {
      "total": {
        "value": 0,
        "relation": "gte"
      },
```
Notice `gte`, not `eq`.
Those 2 errors are connected, as they either don't happen at all (if we have `>0` results in aggregation, everything's fine), or both exist if we get empty results from Clickhouse.


nablaone (approved):
LGTM

PR #1231, opened 2025-01-27
Relaxing explicit index definition override requirement

pdelewski:
Before this change such configuration:
```
        kibana_sample_data_flights:
          tableName: "kibana_sample_data_flights_ext"
```
required also explicit override index definition:
```
        kibana_sample_data_flights_ext:
```
otherwise kibana_sample_data_flights_ext was treaded as non-configured.


pdelewski:
https://github.com/QuesmaOrg/quesma/actions/runs/12987498186/job/36216595220

avelanarius (approved):
LGTM, but an additional integration test would be appreciated (we currently don't have any ITs for table name overrides!)

pdelewski:
> LGTM, but an additional integration test would be appreciated (we currently don't have any ITs for table name overrides!)

Yes, will work on that in next PR

PR #1240, opened 2025-01-31
A/B testing catch-all indexes

nablaone:
A/B testing is enabled by adding two targets. It doesn't work for the catch-all section. Quesma raises a validation error. 
```
        '*':
          useCommonTable: true
          target: [ my-clickhouse-data-source, my-elastic ]
```
This is problematic if we want to A/B test indexes stored in `common_table` and don't want to configure them explicitly. 

This PR adds an option to configure A/B testing for catch-all, by adding a optimizer:
```
        '*':
          useCommonTable: true
          optimizers:
            - elastic_ab_testing:
                disabled: false
```


nablaone:
Dead end.

PR #1246, opened 2025-02-05
Turing-5: Fix escaping in `LIKE` expressions

trzysiek:
`LIKE` pattern string needs different handling, as we need to escape `%` and `_`. They are special characters in `LIKE`, and not in anything else.
Solutions that I saw were new special `InfixExpr` (implemented here), and new `LiteralExpr`. I thought the first option is less intrusive.


quesma/model/expr.go
```diff
@@ -113,6 +113,32 @@ func (e InfixExpr) Accept(v ExprVisitor) interface{} {
 	return v.VisitInfix(e)
 }
 
+type (
```
nablaone, line 116: I'm not sure if we need a special node for the LIKE operator here. It can be represented as an infix node. I'm a big fan of minimal ASTs.

quesma/model/expr_string_renderer.go
```diff
@@ -114,6 +115,39 @@ func (v *renderer) VisitInfix(e InfixExpr) interface{} {
 	}
 }
 
+func (v *renderer) VisitLikeExpr(e LikeExpr) interface{} {
```
nablaone, line 118: Processing on rendering to the string can be problematic. You'll see sth different in the debugger than in logs.

quesma/model/expr.go
```diff
@@ -113,6 +113,32 @@ func (e InfixExpr) Accept(v ExprVisitor) interface{} {
...
+	LikeExpr struct {
+		Left           Expr
+		Op             string
+		Right          Expr
+		BoundType      LikeExprBoundType // Both means we do '%...%' (default), Right means we do '...%' (e.g. for prefix aggregation)
+		AlreadyEscaped bool              // whether the pattern (Right) is already escaped (e.g. % -> \%)
```
nablaone, line 122: This can be replaced with the wrapping type 
```
type AlreadyEscaped struct {
    Value string
}
```
Rendering the literal node should take it into account.

nablaone (commented):
LGTM.



My biggest concern is an architectual one (see comments). Should we add new nodes or keep it minimal as possible. 

@pdelewski please take a look,

trzysiek:
OK, give me a bit to make some changes

nablaone (approved):
LGTM

PR #1248, opened 2025-02-06
Improve validation of Array types

avelanarius:
Before this change, `validateValueAgainstType` would just compare the type name string in case of arrays. This could cause problems in case of arrays of numbers, for example if the user had an `Array(Float64)` ClickHouse column, but tried to insert an array of ints (e.g. `[3, 5, 7]`).

After this fix, the `validateValueAgainstType` logic now works recursively, validating the element type of `Array` (e.g. `Float64` of `Array(Float64)`) and comparing that with an element of incoming array from Elastic API.


mieciu (approved):
awesome! :shipit:

PR #1257, opened 2025-02-10
Quickfix failing e2e

trzysiek:
I run e2e twice now and no errors. It failed because of:
* data race with logger in the test
* sometimes flaky 1 aggregation test, mistype and it had `15` and `25` in 2 places, where the number must be the same (result of the same query)

Sorry for the issue, my PR a few days ago introduced the first error.


quesma/quesma/search_test.go
```diff
@@ -862,7 +861,7 @@ func TestSearchTrackTotalCount(t *testing.T) {
 }
 
 func TestFullQueryTestWIP(t *testing.T) {
-	logger.InitSimpleLoggerForTests()
+	// logger.InitSimpleLoggerForTests()
```
pdelewski, line 864: Should not we remove that?

pdelewski (approved):

PR #1259, opened 2025-02-11
Unify golang type to ClickHouse type conversion in `CREATE TABLE` and validation codepaths, improve `Tuple(...)` validation

avelanarius:
In the ingest processor we had two separate codepaths for converting a Golang type into ClickHouse column type: one used in `CREATE TABLE` codepath (`clickhouse.NewType`) and another in ingest validation (`getTypeName`). The first one had better support for (nested) arrays and maps, which were not fully supported in ingest validation.

Therefore this PR unifies the ingest validation codepath to use the `clickhouse.NewType` from `CREATE TABLE` codepath. As a result, ingest validation now has a better support for nested arrays, maps (see `test_ingest.go`) and the code is now simpler.

### DIFF
--- a/quesma/ingest/ingest_validator.go
+++ b/quesma/ingest/ingest_validator.go
@@ -134,8 +134,9 @@ func validateValueAgainstType(fieldName string, value interface{}, columnType cl
 				}
 				return true
 			}
+		} else {
+			logger.Error().Msgf("MultiValueType validation is not yet supported for type: %v", columnType)
 		}
-		logger.Error().Msgf("MultiValueType validation is not yet supported for type: %v", columnType)
 
 		return false
 	case clickhouse.CompoundType:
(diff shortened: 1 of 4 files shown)


quesma/ingest/ingest_validator.go
```diff
@@ -169,63 +97,66 @@ func validateNumericType(columnType string, incomingValueType string, value inte
...
 			}
+		} else {
+			logger.Error().Msgf("MultiValueType validation is not yet supported for type: %v", columnType)
+		}
 
-			if innerTypesAreCompatible {
-				return validateValueAgainstType(fieldName, valueAsArray[0], compoundType.BaseType)
+		return false
+	case clickhouse.CompoundType:
+		if columnType.Name == "Array" {
```
mieciu, line 143: didn't we have constants for those clickhouse type names? I'd opt for using them in this PR.

mieciu (commented):
One note on using constants for CH type names, but also I don't have a strong opinion, maybe that should land in a separate PR as a greater refactor. 

Other than that LGTM, but I'll defer my approval to someone else :)

quesma/ingest/ingest_validator.go
avelanarius: We don't have any existing constants for these ones (we only have constants for Elastic ones)

mieciu: bummer! yea I'm giving 👍 to whole PR then

mieciu (approved):

pdelewski (approved):

PR #1266, opened 2025-02-13
Add deduplicated log line printing to quesma_logger, use it in places that logged most often

avelanarius:
This PR introduces `logger.DeduplicatedWarn()` and `logger.DeduplicatedInfo()` methods to `quesma_logger`, which automatically deduplicate log messages.

For example,

```
  Ingesting via _bulk API, batch size=2 lines
```

is a log line that the code tries to print very often, but with `DeduplicatedInfo()` that exact log line is only printed once. An expirable LRU cache is used with 1 minute expiry time, so the log line is still printed occasionally, but at most 1 every minute.

Additionally, a couple places that produced a lot of logs were migrated to use `DeduplicatedWarn()`/`DeduplicatedInfo()`.

### DIFF
--- a/quesma/v2/core/quesma_logger.go
+++ b/quesma/v2/core/quesma_logger.go
@@ -24,7 +24,7 @@ const (
 	DefaultBurstSamplerMaxLogsPerSecond = 30  // ~100k lines of logs per hour
 	DefaultSheddingFrequency            = 100 // when the limit is exhausted, log every ~ 100 log lines
 
-  DeduplicatedLogsCacheSize  = 1000
+	DeduplicatedLogsCacheSize  = 1000
 	DeduplicatedLogsExpiryTime = 1 * time.Minute
 )
 
--- a/quesma/v2/core/quesma_logger.go
+++ b/quesma/v2/core/quesma_logger.go
@@ -4,8 +4,11 @@ package quesma_api
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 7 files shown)


quesma/v2/core/quesma_logger.go
```diff
@@ -121,6 +132,45 @@ func (l *QuesmaLoggerImpl) ErrorWithCtx(ctx context.Context) *zerolog.Event {
...
+	// []interface{} is not hashable, so we need to hash it manually
+	// For the convenience sake we just hash a Print representation
+	h := fnv.New32a()
+	fmt.Fprint(h, format, v)
+	return h.Sum32()
+}
+
+// TODO: Add wrappers for other *zerolog.Event methods
+func (m DeduplicatedEvent) Msgf(format string, v ...interface{}) {
+	hash := hashMsgf(format, v)
```
pdelewski, line 164: Non blocking, can we somehow test deduplication logic?

pdelewski:
LGTM :)

pdelewski (approved):

quesma/v2/core/quesma_logger.go
avelanarius: Added a unit test: https://github.com/QuesmaOrg/quesma/pull/1266/commits/197bb87c5789cd359aa65a67c6c45a281f2e9ff4

PR #1267, opened 2025-02-13
Adding support for `ON CLUSTER` clause in DDL statements

mieciu:
This PR adds support for `ON CLUSTER` clause within DDL statements. If user want to synchronize DDL across nodes, it is required to place `clusterName` in the ingest processor configuration like this:
```yaml
  - name: my-ingest-processor
    type: quesma-v1-processor-ingest
    config:
      clusterName: "quesma_cluster"
      indexes:
        example:
```

Having that, Quesma is going to append `ON CLUSTER "quesma_cluster"` to all `CREATE TABLE` and `ALTER TABLE` statements.

###  Some useful bits for testing

**Step 0.** Make sure you configured `clusterName` for table named `try_it`.
  
Test table creation:
```
curl -i -X POST "http://localhost:8080/_bulk" -H "Content-Type: application/ndjson" -d '
{ "index": { "_index": "try_it", "_id": "1" } }
{ "name": "Alice", "age": 30, "city": "New York" }
{ "index": { "_index": "try_it", "_id": "2" } }
{ "name": "Bob", "age": 25, "city": "San Francisco" }
{ "index": { "_index": "try_it", "_id": "3" } }
{ "name": "Charlie", "age": 35, "city": "Los Angeles" }
'
```
Test `ALTER TABLE`:
```
curl -i -X POST "http://localhost:8080/_bulk" -H "Content-Type: application/ndjson" -d '
{ "index": { "_index": "try_it", "_id": "6" } }
{ "name": "Alice", "age": 30, "city": "New York", "genre": "jazz" }
{ "index": { "_index": "try_it", "_id": "8" } }
{ "name": "Bob", "age": 25, "city": "San Francisco", "genre": "hiphop" }
'
```

Also the same steps (create + alter) should be run for common table scenario.

### DIFF
--- a/docs/public/docs/config-primer.md
+++ b/docs/public/docs/config-primer.md
@@ -106,6 +106,7 @@ The supported configuration options for backend connectors (under `config`):
 * `user` - username for authentication
 * `password` - password for authentication 
 * `database` - name of the database to connect to. It is optional for ClickHouse, but strictly required for Hydrolix, where it is also referred as "project".
+* `clusterName` - name of the ClickHouse cluster (optional). This will be used in the `ON CLUSTER clusterName` clause when Quesma creates tables. Setting this option ensures that the created tables are present on all nodes of the distributed cluster.
 * `adminUrl` - URL for administrative operations to render a handy link in Quesma management UI (optional)
 * `disableTLS` - when set to true, disables TLS for the connection (optional)
 
--- a/docs/public/docs/limitations.md
+++ b/docs/public/docs/limitations.md
@@ -24,6 +24,7 @@ Quesma supports only ElasticSearch/Kibana 8.0 or above.
 
 ### ClickHouse limitations
 * When using a cluster deployment of ClickHouse, the tables automatically created by Quesma (during [Ingest](/ingest.md)) will use the `MergeTree` engine. If you wish to use the `ReplicatedMergeTree` engine instead, you will have to create the tables manually with  `ReplicatedMergeTree` engine before ingesting data to Quesma.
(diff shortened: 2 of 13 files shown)


avelanarius:
`clusterName` should be configured on the target level, similarly to existing `tableName`:

```

  - name: my-ingest-processor
    type: quesma-v1-processor-ingest
    config:
      clusterName: "quesma_cluster"
      indexes:
        example:
          target:
            - my-clickhouse-data-source:
                tableName: "asdas"
                clusterName: "quesma_cluster"
```

mieciu:
> `clusterName` should be configured on the target level, similarly to existing `tableName`:
> 
> ```yaml
>   - name: my-ingest-processor
>     type: quesma-v1-processor-ingest
>     config:
>       indexes:
>         example:
>           target:
>             - my-clickhouse-data-source:
>                 tableName: "asdas"
>                 clusterName: "quesma_cluster"
> ```

Yea, so my motivation (except simplicity in implementation obviously) was twofold 
1. In general, one needs to be already connected to a database created with `ON CLUSTER` syntax for this to work.
2. Since we support connecting to one database (which is a parameter of backend connector) I'd highly doubt that user would like to Quesma to create a local table.
Much rather, I'd consider this a mistake/something unwanted, which could potentially end up in unpredictable result (e.g. Quesma getting different table discovery results from different cluster nodes).

So I'm open to discussion, but I feel that more possibilities here could lead to more problems which we could avoid being strict and dead simple.

avelanarius (approved):
LGTM, doing final testing right now

PR #1269, opened 2025-02-14
Extracting cmd module

pdelewski:
This PR is about extracting files needed to build `quesma-cli` executable into separate module.

'/run-integration-tests'

cc @mieciu


nablaone (approved):

mieciu:
This PR did changes to workflow def so we need to trigger job with different params https://github.com/QuesmaOrg/quesma/actions/runs/13651578535

mieciu:
> This PR did changes to workflow def so we need to trigger job with different params https://github.com/QuesmaOrg/quesma/actions/runs/13651578535

This one worked out, let's merge 💪

PR #1282, opened 2025-02-19
Fix creating `Nullable(Tuple)` type

mieciu:
This is an invalid ClickHouse type which can end up with:
```
error executing query: code: 43, message: Nested type Tuple(id Nullable(Int64), ns Nullable(Int64)) cannot be inside Nullable type
```
Tuples, similarly to Arrays, cannot be wrapped in a `Nullable` type. This PR fixes this behavior and adds relevant test.


nablaone (approved):
LGTM

PR #1300, opened 2025-02-25
Getting rid of innermost quesma package

pdelewski:
This PR:
- Removes the innermost `quesma` package.
- Moves inner packages up.
- Moves files like `mappings_test.go` and `highlighter_test.go` to the `elasticsearch` and `elastic_query_dsl` packages, respectively, where they should be located.


nablaone (approved):
LGTM

PR #1303, opened 2025-02-26
Remove `0=0 /* _index = table_name /*` from generated SQL

trzysiek:
Noticed it at the customer https://github.com/QuesmaOrg/quesma/issues/1199

Before, for my test request we generated `WHERE 0=0 /* _index=\'Quesma\' */`
Now we generate `WHERE true`, which is IMO a style improvement.
(I've grepped a lot and I'm pretty sure it's the only place in code where we generate this `0=0` thing)

Ideally I'd write a simple transformation to remove unnecessary `true` from the `WHERE`, but that's more work and I propose to do it in another PR, not very urgently.
(Actually I'll make Claude 3.7 do it, it should manage, shouldn't it?)


nablaone (approved):
LGTM

PR #1305, opened 2025-02-26
Ingest - column name mangling fix

nablaone:
This PR solves a bug. When we have documents like 
```
{"foo": {"10": "1"}}
```
Our column mangling mechanism gives two results, `foo__10` and `foo_10`,  depending on the situation (flatten JSON/unprocessed JSON). This is caused by how we process JSONs which is documented in the `processInsertQuery` function.


pdelewski:
LGTM

pdelewski (approved):

PR #1310, opened 2025-02-27
Small tweaks to NOTICE.MD generator

avelanarius:
1. When `collect-go-dependencies.sh` failed, the "Generate NOTICE.MD" GitHub Actions step didn't fail, because `pipefail` mode wasn't turned on. `shell: bash` fixes that and the step will now correctly fail in such case.

2. `collect-go-dependencies.sh` and `collect-manual-notices.sh` now look for go.mod and NOTICE.MD files recursively (`**` by default doesn't). This catches more things that should be included in `NOTICE.MD`.


pdelewski:
LGTM

PR #1313, opened 2025-02-28
Add tests of lexer dialects (sqlparse, sqlfluff)

avelanarius:
This PR adds tests of our ports of dialects from sqlparse and sqlfluff.

The test files are extracted SQL queries from the tests of sqlparse and sqlfluff. The expected outputs (tokens) are collected by running the queries through the original lexers of sqlparse and sqlfluff.

By doing this, we get a really good test coverage (>10k SQL queries, many edge case/"strange"/"tricky" SQL queries collected) and we can be almost certain that our implementation behaves identically to the original sqlparse/sqlfluff implementations.

The test files are stored in a separate repo (added to Quesma as a Git submodule), because they are large (~10MB).

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -20,6 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: 'true'
 
       - name: Tune GitHub-hosted runner network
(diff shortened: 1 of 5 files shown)


avelanarius:
The `testdata` repository is currently hosted at https://github.com/avelanarius/quesma-testdata-wip, but it should be moved to `QuesmaOrg`.

pdelewski (approved):

PR #1317, opened 2025-03-02
[`sample_logs`] Treat leaf `filter` as metric, not bucket aggregation

trzysiek:
Uncomments 2 tests and fixes 2 last TODO requests from the `sample_logs` dashboard.

Leaf filter is basically a metric, even a very simple one, response always looks like this: `{"doc_count": 12}`. It seems reasonable to me to treat it as a metric.
  
Works, but maybe let me try to refactor the code a bit. Now it's a hackish solution.

Edit: I changed my mind. If we release tomorrow (Wednesday) morning, why not do it with a few more panels on sample dashboards working. Code doesn't look that bad to me and we can polish it afterwards. That's the version I showed today - almost everything works on all samples.


mieciu (approved):
:shipit:

PR #1320, opened 2025-03-02
[`sample_ecommerce`] Fix group by SQL for arrays

trzysiek:
Before (Quesma):
Screenshot 2025-03-02 at 16 25 24
After, we give the same results as Elastic:
Screenshot 2025-03-02 at 16 25 39


jakozaur (approved):

jakozaur:
More tests would be nice.

PR #1321, opened 2025-03-02
[all `sample`] Fix Elastic's `.keyword`, `.text`

trzysiek:
It's probably not the perfect solution, as e.g. we might enable search on some type of column which will be slow.
It's not a huge issue for now though, I guess.
Before:
Screenshot 2025-03-02 at 16 28 39
After:
Screenshot 2025-03-02 at 16 29 39


platform/functionality/field_capabilities/field_caps.go
```diff
@@ -70,9 +70,8 @@ func handleFieldCapsIndex(cfg map[string]config.IndexConfiguration, schemaRegist
 			for fieldName, field := range fieldsWithAliases {
 				addFieldCapabilityFromSchemaRegistry(fields, fieldName.AsString(), field.Type, resolvedIndex)
 				switch field.Type.Name {
-				case "text":
+				case "keyword", "text":
 					addFieldCapabilityFromSchemaRegistry(fields, fmt.Sprintf("%s%s", fieldName.AsString(), types.MultifieldKeywordSuffix), schema.QuesmaTypeKeyword, resolvedIndex)
-				case "keyword":
 					addFieldCapabilityFromSchemaRegistry(fields, fmt.Sprintf("%s%s", fieldName.AsString(), types.MultifieldTextSuffix), schema.QuesmaTypeText, resolvedIndex)
```
jakozaur, line 75: I don't understand what we do here, it looks like we always add those suffices.

jakozaur (approved):
MErgable if it fixes samples.

platform/functionality/field_capabilities/field_caps.go
trzysiek: Yeah, and before we were adding `.keyword` suffix to text fields, and `.text` to keywords... After this PR we only change, that for keywords we add `.keyword`, and for texts `.text`. So this PR isn't even sketchy, I think. Our previous solution was.

PR #1323, opened 2025-03-02
[Grafana] Handle `min_doc_count` also in `terms`

trzysiek:
I do that by adding `count() >= min_doc_count DESC` to the `ORDER BY` (at the beginning). Then we'll have rows with `>= min_doc_count` first, and can filter out the rest.

Before:
Screenshot 2025-03-02 at 21 16 36
After:
Screenshot 2025-03-02 at 21 15 40


platform/model/bucket_aggregations/terms.go
```diff
@@ -245,3 +250,27 @@ func CheckParamsTerms(ctx context.Context, paramsRaw any) error {
...
+
+func (query Terms) NewRowsTransformer() model.QueryRowsTransformer {
+	return &TermsRowsTransformer{minDocCount: int64(query.minDocCount)}
+}
+
+type TermsRowsTransformer struct {
+	minDocCount int64
+}
+
+func (qt TermsRowsTransformer) Transform(ctx context.Context, rowsFromDB []model.QueryResultRow) []model.QueryResultRow {
```
jakozaur, line 262: I saw this code before, any chance to unify that logic.

platform/parsers/elastic_query_dsl/pancake_aggregation_parser_buckets.go
```diff
@@ -178,18 +184,24 @@ func (cw *ClickhouseQueryTranslator) parseTermsAggregation(aggregation *pancakeA
...
 	}
 
 	aggregation.queryType = terms
 	aggregation.selectedColumns = append(aggregation.selectedColumns, field)
-	aggregation.limit = size
+	aggregation.limit = cw.parseSize(params, defaultSize)
 	aggregation.orderBy = orderBy
+	if minDocCount > 1 {
+		// need to filter out buckets with doc_count < min_doc_count, so first order by is (count() >= min_doc_count) DESC
+		condition := model.NewInfixExpr(model.NewCountFunc(), ">=", model.NewLiteral(minDocCount))
```
jakozaur, line 198: I don't understand why we do that here. Probably would love to talk about this logic. Previously similar filtering was done on client side and that kind of logic was not neccessary.

jakozaur: Ok, please add just better comment with example.

platform/model/bucket_aggregations/terms.go
trzysiek: Good idea, adding as TODO, as merging this PR asap would help me in some other ones.

jakozaur (approved):

PR #1327, opened 2025-03-02
[Grafana] `geohash_grid`: Support the only new aggregration from Grafana

trzysiek:
Doesn't support `bounds` parameter, but will warn to logs if someone uses it. Only `precision` is visible in Grafana, so doesn't really matter there. Let's add that in another PR if needed.

### DIFF
--- a/platform/model/bucket_aggregations/geohash_grid.go
+++ b/platform/model/bucket_aggregations/geohash_grid.go
@@ -84,7 +84,7 @@ func CheckParamsGeohashGrid(ctx context.Context, paramsRaw any) error {
 		if _, isRequired := requiredParams[paramName]; !isRequired {
 			wantedType, isOptional := optionalParams[paramName]
 			if !isOptional {
-				return fmt.Errorf("unexpected parameter %s found in IP Range params %v", paramName, params)
+				return fmt.Errorf("unexpected parameter %s found in GeoHash Grid params %v", paramName, params)
 			}
 			if reflect.TypeOf(params[paramName]).Name() != wantedType { // TODO I'll make a small rewrite to not use reflect here
 				return fmt.Errorf("optional parameter %s is not of type %s, but %T", paramName, wantedType, params[paramName])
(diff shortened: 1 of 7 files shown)


platform/model/bucket_aggregations/geohash_grid.go
```diff
@@ -0,0 +1,103 @@
...
+			return fmt.Errorf("required parameter %s is not of type %s, but %T", paramName, paramType, paramVal)
+		}
+	}
+
+	// check if only required/optional are present
+	for paramName := range params {
+		if _, isRequired := requiredParams[paramName]; !isRequired {
+			wantedType, isOptional := optionalParams[paramName]
+			if !isOptional {
+				return fmt.Errorf("unexpected parameter %s found in IP Range params %v", paramName, params)
```
jakozaur, line 87:
```suggestion
				return fmt.Errorf("unexpected parameter %s found in IGeohash Grid params %v", paramName, params)
```

platform/model/bucket_aggregations/geohash_grid.go
```diff
@@ -0,0 +1,103 @@
...
+			if reflect.TypeOf(params[paramName]).Name() != wantedType { // TODO I'll make a small rewrite to not use reflect here
+				return fmt.Errorf("optional parameter %s is not of type %s, but %T", paramName, wantedType, params[paramName])
+			}
+		}
+	}
+
+	// log if you see them
+	for _, warnParam := range logIfYouSeeThemParams {
+		if _, exists := params[warnParam]; exists {
+			logger.WarnWithCtxAndThrottling(ctx, "ip_prefix", warnParam, "we didn't expect %s in IP Range params %v", warnParam, params)
```
jakozaur, line 98:
```suggestion
			logger.WarnWithCtxAndThrottling(ctx, "ip_prefix", warnParam, "we didn't expect %s in Geohash Grid params %v", warnParam, params)
```

platform/model/bucket_aggregations/geohash_grid.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	logIfYouSeeThemParams := []string{"bounds", "shard_size"}
+
+	params, ok := paramsRaw.(model.JsonMap)
+	if !ok {
+		return fmt.Errorf("params is not a map, but %+v", paramsRaw)
+	}
+
+	// check if required are present
+	for paramName, paramType := range requiredParams {
```
jakozaur, line 72: Maybe this should be function re-used elsewhere.

jakozaur (approved):
Looks good, but some minor changes are required.

PR #1329, opened 2025-03-02
[`sample_flights`, quickfix] Add `doc_count = 0` if no DB result in `filter`

trzysiek:
Also adds all requests from `flights` as tests. I think they all pass, but I need to update results in last 2 of them.


platform/parsers/elastic_query_dsl/pancake_sql_query_generation_test.go
```diff
@@ -48,6 +48,14 @@ func TestPancakeQueryGeneration(t *testing.T) {
 
 	for i, test := range allAggregationTests() {
 		t.Run(test.TestName+"("+strconv.Itoa(i)+")", func(t *testing.T) {
+			// sample_flights
+			if test.TestName == "Delays & Cancellations (request 1/2)(file:kibana-sample-data-flights,nr:2)" {
+				t.Skip()
```
jakozaur, line 53: Would add some comment if it's just test or real issue.

platform/parsers/elastic_query_dsl/pancake_sql_query_generation_test.go
```diff
@@ -48,6 +48,14 @@ func TestPancakeQueryGeneration(t *testing.T) {
...
+			}
+			if test.TestName == "TODO Airport Connections (Hover Over Airport)(file:kibana-sample-data-flights,nr:14)" {
+				t.Skip()
```
jakozaur, line 56: Same here.

jakozaur (approved):
Please just add two comments.

PR #1344, opened 2025-03-04
Field map syntax 

nablaone:
- **Add support for accessing map values by key**
- **Add support for Map(LowCardinality(String),Strign) type**
- **Opt-in flag**


pdelewski (approved):

jakozaur:
@mieciu @nablaone @pdelewski  For now, it is an experimental feature. We should design later a proper way to pass extra options.

pdelewski:
> @mieciu @nablaone @pdelewski For now, it is an experimental feature. We should design later a proper way to pass extra options.

Added issue for tracking that https://github.com/QuesmaOrg/quesma/issues/1351

PR #1345, opened 2025-03-04
Add `PARTITION BY` to index config 

trzysiek:
Maybe it's enough? Not sure here though, I don't have much experience with ingest.


mieciu:
As discussed on Zoom, we'll be moving forwards with this approach 👉 https://github.com/QuesmaOrg/quesma/pull/1347

PR #1347, opened 2025-03-04
Adding experimental (and limited) support for table partitioning

mieciu:
We're adding a `partitioningStrategy` to index configuration:
```yaml
processors:
  - name: my-query-processor
    type: quesma-v1-processor-query
    config:
      indexes:
        '*':
          partitioningStrategy: "daily"
        try_it:
          partitioningStrategy: "monthly"
```
Having `partitioningStrategy` essentially adds `PARTITION BY function(` to `CREATE TABLE` statement executed by Quesma.

Key points:
1) `partitioningStrategy` is applied only on table creation handled by Quesma. **If the table already exists, Quesma won't modify it.**
2) `partitioningStrategy` can be set to `hourly`, `daily`, `monthly` and `yearly`.
3) **We're not adding common table partitioning support** - so if index has `useCommonTable: true` specified together with any `partitioningStrategy` - Quesma is going to fail due to invalid configuration
4) Leaving `partitioningStrategy` not specified means table is not partitioned at all (== ClickHouse default)


platform/clickhouse/clickhouse.go
```diff
@@ -55,10 +55,11 @@ type (
 		ClusterName          string // Name of the cluster if created with `CREATE TABLE ... ON CLUSTER ClusterName`
 		Engine               string // "Log", "MergeTree", etc.
 		OrderBy              string // "" if none
-		PartitionBy          string // "" if none
-		PrimaryKey           string // "" if none
-		Settings             string // "" if none
-		Ttl                  string // of type Interval, e.g. 3 MONTH, 1 YEAR
+		//PartitionBy          string // "" if none
```
pdelewski, line 58: I think we remove that and below

platform/config/config_v2.go
```diff
@@ -341,6 +341,16 @@ func (c *QuesmaNewConfiguration) validatePipelines() error {
 				if queryIndexConf.UseCommonTable != ingestIndexConf.UseCommonTable {
 					return fmt.Errorf("ingest and query processors must have the same configuration of 'useCommonTable' for index '%s' due to current limitations", indexName)
 				}
+				if queryIndexConf.PartitioningStrategy != ingestIndexConf.PartitioningStrategy {
+					return fmt.Errorf("ingest and query processors must have the same configuration of 'partitioningStrategy' for index '%s' due to current limitations", indexName)
+				}
+				if ingestIndexConf.PartitioningStrategy != "" && ingestIndexConf.UseCommonTable { // we don't check for queryIndexConf.PartitioningStrategy per previous check
```
pdelewski, line 347: Should not we also have similar check for `queryIndexConf`?

mieciu: https://github.com/QuesmaOrg/quesma/pull/1347/commits/3bdef66a7cb6c7b0fdb87e4d9746524c06040e6e

pdelewski (approved):

PR #1348, opened 2025-03-04
Take local platform dependency

pdelewski:


mieciu:
Failure in that triggered IT run was expected since it has used job definitions from `main`, here's a "legally binding" run https://github.com/QuesmaOrg/quesma/actions/runs/13675389549 😉

mieciu (approved):
I approve.

PR #1355, opened 2025-03-06
Nested type resolution fixes for `table_discovery.go`

mieciu:
It turned out that we can't handle double-nested arrays in ClickHouse, which caused tables not being visible in Quesma during table discovery.


nablaone (approved):
LGTM

pdelewski:
> LGTM

For me too :)

PR #1356, opened 2025-03-07
Revert full failed insert statement logging

mieciu:


pdelewski:
Just to confirm, is this due log spamming?

mieciu:
> Just to confirm, is this due log spamming?

Correct, single line here can easily get to >1MB in size

PR #1363, opened 2025-03-12
Support Kibana CSV export

mieciu:
Adding a support for CSV export in Kibana.

image

### Some technical details
Kibana CSV report generation uses two steps:
1. Create Point-In-Time  (`:indexPattern/_pit` API) for an index pattern
2. Fetch data via `/_search` API, referencing PIT ID created before

In our case, we're **not** creating any persistent PIT for ClickHouse table - we're simply using that PIT ID as a reference to specific index pattern. So the logic looks like the following:
1. Quesma controls `:indexPattern/_pit` endpoint.
  Whenever pit creation is issued and `:indexPattern` is under our control, instead of relaying request to Elasticsearch we respond that PIT has been created and it's ID is `quesma_:indexPattern`.
2. When a request to `/_search` comes to Quesma and it has PIT == `quesma_:indexPattern`, we simply execute query to a table which is matching `:indexPattern`

Having that, we have this feature implemented without the need of any persistent storage 🎉

### DIFF
--- a/platform/frontend_connectors/matchers.go
+++ b/platform/frontend_connectors/matchers.go
@@ -45,6 +45,57 @@ func matchAgainstTableResolver(indexRegistry table_resolver.TableResolver, pipel
 	})
 }
 
+func isNotTargetingInternalIndex() quesma_api.RequestMatcher {
+	return quesma_api.RequestMatcherFunc(func(req *quesma_api.Request) quesma_api.MatchResult {
+		indexPattern := req.Params["indexPattern"]
+		if strings.HasPrefix(indexPattern, ".") {
+			return quesma_api.MatchResult{Matched: false}
+		} else {
+			return quesma_api.MatchResult{Matched: true}
+		}
+	})
+}
+
+// getPitIdFromRequest gets the PIT ID from the request body,
+// depending on request kind it can be either at root or under `pit` key, e.g.:
+// {"id": "pit_id"} or {"pit": {"id": "pit_id"}}
+func getPitIdFromRequest(req *quesma_api.Request, pitAtRoot bool) string {
+	var payload struct {
+		ID  string `json:"id,omitempty"`
+		Pit struct {
+			ID string `json:"id"`
+		} `json:"pit,omitempty"`
+	}
+	if err := json.Unmarshal([]byte(req.Body), &payload); err != nil {
+		return ""
+	}
+	if pitAtRoot {
+		return payload.ID
+	}
+	return payload.Pit.ID
+}
+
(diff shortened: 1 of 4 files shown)


platform/frontend_connectors/router_v2.go
```diff
@@ -133,22 +133,31 @@ func ConfigureSearchRouterV2(cfg *config.QuesmaConfiguration, dependencies quesm
 		return HandleResolveIndex(ctx, req.Params["index"], sr, cfg.Elasticsearch)
 	})
 
+	router.Register(routes.IndexPatternPitPath, and(method("POST"), isNotTargetingInternalIndex(), matchedAgainstPattern(tableResolver)), func(ctx context.Context, req *quesma_api.Request, _ http.ResponseWriter) (*quesma_api.Result, error) {
```
avelanarius, line 136: `isNotTargetingInternalIndex` and `matchedAgainstPattern` are really similar, but `isNotTargetingInternalIndex` doesn't use table resolver.

I think that you should make a variant of `matchedAgainstPattern` that extracts the index name from `req.Params["indexPattern"]` - that way the code will use table resolver, instead of a new codepath.

pdelewski (approved):

platform/frontend_connectors/router_v2.go
pdelewski: Good point!. Would be good to use `matchedAgainstPattern` if that's possible

mieciu: yea that has been really tempting, though I really liked the verbosity in the router code and how explicit the matching rules look - so decided to keep this separate.

I don't have a strong preference either way - so knowing my stance feel free to decide :)

platform/frontend_connectors/matchers.go
```diff
@@ -21,17 +21,18 @@ func matchedAgainstAsyncId() quesma_api.RequestMatcher {
...
-// check whether exact index name is enabled
 func matchAgainstTableResolver(indexRegistry table_resolver.TableResolver, pipelineName string) quesma_api.RequestMatcher {
 	return quesma_api.RequestMatcherFunc(func(req *quesma_api.Request) quesma_api.MatchResult {
-
 		indexName := req.Params["index"]
 
+		if strings.HasPrefix(indexName, ".") { // Skip internal indices
+			return quesma_api.MatchResult{Matched: false}
+		}
+
```
avelanarius, line 35: Nit: table resolver is already doing that at the start with this first rule:

```
{"kibanaInternal", resolveInternalElasticName},
```

avelanarius (approved):
LGTM, but left one additional nit.

PR #1364, opened 2025-03-12
Introducing map keys discovering

pdelewski:
This PR adds an automatic map key discovery feature needed for field caps and dashboard creation.

```
CREATE TABLE IF NOT EXISTS "foo"
(
    "@timestamp" DateTime64 DEFAULT now64(),
    `bar` Map(String, Nullable(String))
)
ENGINE = MergeTree
ORDER BY ("@timestamp");    

insert into foo values ('2020-01-01 00:00:00', {'a': 'b'});
insert into foo values ('2020-01-01 00:00:00', {'c': 'b'});
insert into foo values ('2020-01-01 00:00:00', {'e': 'b'});
```

Feature is hidden behind the global flag:

```
mapFieldsDiscoveringEnabled: true
```

image

### DIFF
--- a/platform/clickhouse/schema.go
+++ b/platform/clickhouse/schema.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/QuesmaOrg/quesma/platform/config"
 	"github.com/QuesmaOrg/quesma/platform/logger"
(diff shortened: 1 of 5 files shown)


mieciu (approved):
looks good to me!

PR #1367, opened 2025-03-13
Removing field caps map capabilities to be on par with elasticsearch

pdelewski:
Elasticsearch does not emit `.keys()` and `.values()` for maps. Definitely we should not emit `text` type as well. Below `bar` index is handled by elasticsearch.

image

and `foo` index handled by `quesma` and `clickhouse` after fix.

image


mieciu (approved):
👍

PR #1368, opened 2025-03-14
Map fields discovery fix, additional logging

pdelewski:
This PR fixes the addition of a map field to the schema in case of subsequent query failures and adds more logging.

### DIFF
--- a/platform/clickhouse/table_discovery.go
+++ b/platform/clickhouse/table_discovery.go
@@ -664,6 +664,7 @@ func (td *tableDiscovery) enrichTableWithMapFields(inputTable map[string]map[str
 					var key string
 					if err := rows.Scan(&key); err != nil {
 						logger.Debug().Msgf("Error scanning key for table, column: %s, %s, %v", table, colName, err)
+						rowsNext = rows.Next()
 						continue
 					}
 					// Add virtual column for each key in the map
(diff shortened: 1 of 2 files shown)


platform/clickhouse/table_discovery.go
```diff
@@ -631,40 +631,64 @@ func (td *tableDiscovery) enrichTableWithMapFields(inputTable map[string]map[str
...
+					} else {
+						outputTable[table][mapKeyCol] = columnMetadata{
 							colType: valueType,
 							origin:  schema.FieldSourceMapping,
 						}
+						logger.Debug().Msgf("Added map key column: %s.%s", table, mapKeyCol)
 					}
+					rowsNext = rows.Next()
+				}
+				err = rows.Close() // Close after processing
```
nablaone, line 688: There should be a check if rows.Err() is not nil after iteration and before close.

platform/clickhouse/table_discovery.go
```diff
@@ -631,40 +631,64 @@ func (td *tableDiscovery) enrichTableWithMapFields(inputTable map[string]map[str
...
 
-				// Ensure the table exists in outputTable
-				if _, ok := outputTable[table]; !ok {
-					outputTable[table] = make(map[string]columnMetadata)
-				}
-
 				// Process returned keys and add them as virtual columns
-				for rows.Next() {
+				rowsNext := rows.Next()
+				if !rowsNext {
```
nablaone, line 655: The  same here.

nablaone (changes requested):

platform/clickhouse/table_discovery.go
pdelewski: Made little rewrite. Check now

nablaone (approved):

PR #1366, opened 2025-03-13
[`ecommerce`] Fix failing panels when filtering an `Array(String)`

trzysiek:
I think code is clearer now, and of course some bugs also seem fixed.

Before: 3 panels started failing when I choose any of the array filters, as I discovered and showed you on last learn.
Screenshot 2025-03-13 at 11 58 52.
Screenshot 2025-03-13 at 11 58 57

After: they all load again. 
Screenshot 2025-03-13 at 11 59 40

PR #1373, opened 2025-03-16
[`ecommerce 2/2`] Make string pattern matching more Elastic-like

trzysiek:
~Merge https://github.com/QuesmaOrg/quesma/pull/1366 first please, it'll be easier for me - they both touch similar things~

Before: Elastic matched only `Men's clothing`. We also match `Women's clothing`:
Screenshot 2025-03-16 at 09 17 01
Screenshot 2025-03-16 at 09 17 06


pdelewski:
Too many conflicts

Issue #1380, opened 2025-03-25
data type convert problem


marksmansj:
Hello. All:
there is a timestamp column in my clickhouse . its data type is Uint64 and stores milliseconds unix timestamp. when I use kibana to create data view. kibana can not identify the timestamp . the timestamp type is integer in kibana data view. does anyone know how to solve this problem?  thank you All.
![Image](https://github.com/user-attachments/assets/c46c1df7-3e33-4a8e-a507-e6471f51ce16)

![Image](https://github.com/user-attachments/assets/fe5cc7de-667e-46a8-8a0c-556f8bf918ee)

trzysiek:
Hey, yeah, unfortunately it's not a frequent case and it didn't work for me out of the box either, but I managed to make it work. I'll need to prepare a small code update for that to work reliable, but should be ready today in the evening. I'll post you exactly what to do by then.

pdelewski:
Work in progress

PR #1381, opened 2025-03-25
Support `timestamp` in Kibana being `int` (not `DateTime`) column in ClickHouse

trzysiek:
Basically support for https://github.com/QuesmaOrg/quesma/issues/1380

### DIFF
--- a/platform/frontend_connectors/schema_transformer_test.go
+++ b/platform/frontend_connectors/schema_transformer_test.go
@@ -19,6 +19,10 @@ type fixedTableProvider struct {
 	tables map[string]schema.Table
 }
 
+func newFixedTableProvider(tables map[string]schema.Table) *fixedTableProvider {
+	return &fixedTableProvider{tables: tables}
+}
+
 func (f fixedTableProvider) TableDefinitions() map[string]schema.Table {
 	return f.tables
 }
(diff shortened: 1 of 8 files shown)


platform/frontend_connectors/schema_transformer_test.go
```diff
@@ -1600,3 +1586,247 @@ func Test_mapKeys(t *testing.T) {
 	}
 
 }
+
+func Test_acceptIntsAsTimestamps(t *testing.T) {
+	schemaTable := schema.Table{
+		Columns: map[string]schema.Column{
+			"@timestamp":   {Name: "@timestamp", Type: "DateTime64"},
+			"timestampInt": {Name: "timestampInt", Type: "UInt64"},
```
pdelewski, line 1594: As we discussed, I think this should be `DateTime`, however it would be good to check how elasticsearch behaves in this case.

trzysiek: I'll check that

pdelewski:
I created following table 
```
CREATE TABLE cat
(
    timestamp UInt64,
    name String,
    age UInt32
)
ENGINE = MergeTree
PARTITION BY toYYYYMM(toDateTime(timestamp))
ORDER BY (timestamp);
```
and emitted few inserts
```
INSERT INTO cat (timestamp, name, age) VALUES
(1715126400, 'Whiskers', 3),
(1715212800, 'Mittens', 5),
(1715299200, 'Shadow', 2),
(1715385600, 'Luna', 4),
(1715472000, 'Simba', 1);
```

then extended config with:
```
        cat:
          schemaOverrides:
            fields:
              "timestamp":
                type: "date"
```

but it seems that it does not return results, now question is whether it's bug or type for timestamp should be different

pdelewski:
It seems that there is something wrong with timestamps as query looks good and returns result when timestamp is correct
image

pdelewski:
According to elasticsearch types, `date` is correct in this context

trzysiek:
Sorry it took me so long, but you know I had some other priorities. I should do it very soon, looks like something trivial.

trzysiek:
Amazing walkthrough ofc, sorry it was necessary.

pdelewski (approved):

PR #1383, opened 2025-03-28
Small fixes for new experimental maps

trzysiek:


platform/functionality/terms_enum/terms_enum.go
```diff
@@ -92,16 +95,23 @@ func handleTermsEnumRequest(ctx context.Context, body types.JSON, lm clickhouse.
...
 	where := qt.ParseAutocomplete(indexFilter, field, prefixString, caseInsensitive)
 	selectQuery := buildAutocompleteQuery(field, qt.Table.Name, where.WhereClause, size)
-	dbQueryCtx, cancel := context.WithCancel(ctx)
-	// TODO this will be used to cancel goroutine that is executing the query
-	_ = cancel
-
-	if rows, _, err2 := lm.ProcessQuery(dbQueryCtx, qt.Table, selectQuery); err2 != nil {
-		logger.Error().Msgf("terms enum failed - error processing SQL query [%s]", err2)
-		result, err = json.Marshal(emptyTermsEnumResponse())
+	selectQuery, err = transformations.ApplyAllNecessaryTransformations(selectQuery, qt.Schema, isFieldMapSyntaxEnabled)
```
pdelewski, line 98: Now at least this will be consistent regarding transformations

platform/transformations/necessary_transformations.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"github.com/QuesmaOrg/quesma/platform/common_table"
+	"github.com/QuesmaOrg/quesma/platform/model"
+	"github.com/QuesmaOrg/quesma/platform/schema"
+	"strings"
+)
+
+// TODO it should be removed and transformations should be in 1 place.
```
pdelewski, line 13: Agree with that. In fact handling terms enum should be rewritten and combined with other queries. That's just another query.

pdelewski (approved):

platform/transformations/necessary_transformations.go
trzysiek: Thanks for the remainder. I'll try to do that pretty soon. It seems more important than at least some of my ideas for gradual tech debt removal.

PR #1386, opened 2025-03-29
[done, needs some testing] Revert `Text/keyword` change?

trzysiek:
Before this PR all sample dashboards load in Kibana 8.11, but one single panel doesn't in newest Kibana 8.17.
After this PR they all load in both of them.

https://github.com/QuesmaOrg/quesma/pull/1321

Draft for a bit, let me inspect a) PR above b) the customer case which was why I also introduced the reverse change a few months ago (we used to have exactly the same version in code as after this PR months ago)

### DIFF
--- a/platform/functionality/field_capabilities/field_caps.go
+++ b/platform/functionality/field_capabilities/field_caps.go
@@ -70,8 +70,9 @@ func handleFieldCapsIndex(cfg map[string]config.IndexConfiguration, schemaRegist
 			for fieldName, field := range fieldsWithAliases {
 				addFieldCapabilityFromSchemaRegistry(fields, fieldName.AsString(), field.Type, resolvedIndex)
 				switch field.Type.Name {
-				case "keyword", "text":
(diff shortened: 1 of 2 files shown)


platform/functionality/field_capabilities/field_caps.go
```diff
@@ -70,8 +70,9 @@ func handleFieldCapsIndex(cfg map[string]config.IndexConfiguration, schemaRegist
 			for fieldName, field := range fieldsWithAliases {
 				addFieldCapabilityFromSchemaRegistry(fields, fieldName.AsString(), field.Type, resolvedIndex)
 				switch field.Type.Name {
-				case "keyword", "text":
+				case "text":
```
pdelewski, line 73: Seems like "text", should be swapped with "keyword", at least looking at capabilities we are adding

pdelewski:
I don't want to block this, however for the future we should have failing test for `Kibana 8.17` first

Issue #1377, opened 2025-03-20
does quesma support multiple clickhouse databases in backendConnectors


marksmansj:
Hello. All:
in my sceanario. there are multiple databases in clickhouse , such as ops/dev . how can i expose tables in these databases ? can backendConnectors config multiple databases ?

trzysiek:
aa

marksmansj:
> Hi [@marksmansj](https://github.com/marksmansj) ! Are those databases administered by the same clickhouse server? If yes, then multiple databases support is in Quesma out of the box.
> 
> I can show you full walk-through how to make it work. I already did, but later thought it's probably to ask the question above first.

thank you . yes. these databases are in the same clickhouse cluster. by the way. do you know how to get a licenseKey? I've send a email to support@quesma.com to ask a key . but no one reponse .

trzysiek:
Hi. I think I can say we support multiple databases on the same cluster. Let me show you a quick example how I achieved what I think you're trying to achieve, in just a few minutes.

I run a default Quesma stack, so Quesma + Clickhouse + Kibana (with Quesma config that I paste at the bottom), and first connected to Clickhouse and set up a couple of databases there:
![Image](https://github.com/user-attachments/assets/fb9a8d16-9b47-4280-b963-ee3cfe2e2c36)
Created a few tables on a few different databases, e.g.:
![Image](https://github.com/user-attachments/assets/ded3b375-f4cf-4f1f-8a37-9e9ccdae7d3c)
![Image](https://github.com/user-attachments/assets/dc055322-4a57-4b55-9b2a-1615e17aeb9f)
Inserted a few example values to the tables, then opened Kibana, Explorer > Data Views and you can see there is some data from `default.kibana_sample_data_logs` table.

Then I clicked the stuff below to create a Data View in Kibana to explore my data and create all the dashboards that Kibana supports.
![Image](https://github.com/user-attachments/assets/752b65f1-bc23-4d0f-aa8f-0744a2d8b570)
![Image](https://github.com/user-attachments/assets/f21d3f15-f4b3-4a7a-9f4e-aa51fdeca69a)

And finally I see all the data from `dev.d` table. I also see tables from databases `default` and `q`.

My minimalistic Quesma config:
```
disableTelemetry: true
frontendConnectors:
  - name: elastic-ingest
    type: elasticsearch-fe-ingest
    config:
      listenPort: 8080
      disableAuth: true
  - name: elastic-query
[...]

PR #1390, opened 2025-03-30
Allow experimental support for `>1` backend connector

trzysiek:
Support for customer issue https://github.com/QuesmaOrg/quesma/issues/1377

Before we simply didn't allow `>1` backend connector. I think some basic dropping of this limitation might not be too problematic, but not 100% sure.


pdelewski:
I think that problem described in https://github.com/QuesmaOrg/quesma/issues/1377 is different. It's about supporting/querying tables from more than one database at the same time.
So, one option would be to support such kind of configuration:
```
 - name: clickhouse-instance
    type: clickhouse-os
    config:
      url: clickhouse://192.168.60.211:9000
      user: admin
      password: 12345
      database: [ops, dev]
```

Other option would be to discover tables from all databases when `database` is not defined. At the moment we refer to `default` by default.

pdelewski:
I'm closing this as preferred way to fix it would be the one described by above comment

PR #1392, opened 2025-03-31
[Transpiler] Small leftovers of refactor

trzysiek:
change back to `Add(pipe, ...nodes)` from `Add(&pipe, ...nodes)` which I like more, but idk, maybe you don't.

@avelanarius please look and reject/merge whenever you like, we can wait with this of course.


pdelewski:
LGTM. @avelanarius Can we merge it?

Issue #1394, opened 2025-04-06
Quesma authentication


hsuanyi:
from quesma to opensearch/elasticsearch, does quesma only support basic auth? is there a plan to support mtls?

pdelewski:
Fixed by #1406

PR #1406, opened 2025-04-25
Support Client certificate authentication for opensearch/elastic search 

hsuanyi:
In addition to username and password, es/os support client-auth based auth. See [here](https://docs.opensearch.org/docs/latest/security/authentication-backends/client-auth/). This pr wants to support that. 

This change introduces client certificate authentication for Quesm when connecting to Elasticsearch. It does not alter how users authenticate to Quesm.

(this pr is to implement request of this https://github.com/QuesmaOrg/quesma/issues/1394)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ FROM golang:alpine AS builder
 COPY platform /platform
 COPY cmd /cmd
 
+ENV GOOS=linux
+ENV GOARCH=amd64
 ENV GOCACHE=/root/.cache/go-build
 ARG QUESMA_BUILD_SHA
 ARG QUESMA_VERSION
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,6 @@ FROM golang:alpine AS builder
 COPY platform /platform
 COPY cmd /cmd
 
-ENV GOOS=linux
-ENV GOARCH=amd64
 ENV GOCACHE=/root/.cache/go-build
 ARG QUESMA_BUILD_SHA
 ARG QUESMA_VERSION
--- a/platform/elasticsearch/client.go
+++ b/platform/elasticsearch/client.go
@@ -27,35 +27,33 @@ type SimpleClient struct {
 }
 
 func NewHttpsClient(configuration *config.ElasticsearchConfiguration, timeout time.Duration) *http.Client {
-	var cert = tls.Certificate{}
-	if configuration.ClientCertPath != "" && configuration.ClientKeyPath != "" {
-		var err error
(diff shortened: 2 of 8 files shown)


pdelewski:
Huge thanks to @hsuanyi  for jumping in here — really appreciate you taking the time to help move the project forward!

hsuanyi:
@pdelewski it is ready for review. Thanks!

also this is for this issue: https://github.com/QuesmaOrg/quesma/issues/1394

platform/telemetry/phone_home.go
```diff
@@ -87,7 +87,8 @@ type agent struct {
 	recent            diag.PhoneHomeStats
 	telemetryEndpoint *config.Url
 
-	httpClient *http.Client
+	httpClient              *http.Client
```
mieciu, line 90: I don't think this `httpClient` is needed anymore

Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM golang:alpine AS builder
 COPY platform /platform
 COPY cmd /cmd
 
+ENV GOOS=linux
+ENV GOARCH=amd64
```
mieciu, line 7: We shouldn't need these variables

platform/elasticsearch/client.go
```diff
@@ -24,18 +26,50 @@ type SimpleClient struct {
...
+func NewHttpsClient(configuration *config.ElasticsearchConfiguration, timeout time.Duration) *http.Client {
+	tlsConfig := &tls.Config{
+		MinVersion:         tls.VersionTLS12,
+		InsecureSkipVerify: true,
+	}
+
+	if configuration.CACertPath != "" {
+		caCert, err := os.ReadFile(configuration.CACertPath)
+		if err != nil {
+			panic(fmt.Sprintf("failed to read CA certificate: %v", err))
```
mieciu, line 38: We don't want `panic`s in production code. 
I think logging at `Error` level and fallback to certificate-less client would be more appropriate.

After all, the server will still throw handshake error, yet at least the service won't crash or bootloop in Cloud/k8s-based environments.

mieciu (changes requested):
@hsuanyi Thank you for your contribution, it's really appreciated !! 🏅 

I got some concerns w.r.t. your changes, please have a look :)

platform/elasticsearch/client.go
hsuanyi: done.

hsuanyi:
@mieciu Addressed. Would you mind taking another look?

mieciu (approved):
I just added a note for future self which should have a look at redundant `SimpleClient` which doesn't make sense anymore in the comment.

Other than that, LGTM, we really appreciate your contribution!

PR #1421, opened 2025-05-12
Send slack message on new issue

pdelewski:


.github/workflows/issues-and-prs-yml
```diff
@@ -0,0 +1,29 @@
+name: Issues and PRs Notification
+
+permissions:
```
pdelewski, line 3: Not sure if that's permissions are needed

PR #1435, opened 2025-06-02
Introduce unique ids

mieciu:
An attempt to introduce unique object IDs in the ClickHouse realm.

A returned document id (`_id` field) would have the following syntax:
```
{hex-encoded timestamp field}qqq{hex-encoded hash of the document}
```
Of course, the `{hex-encoded hash of the document}` lives only in Quesma memory: 
1. It needs to be re-calculated when returning search hits
2. Quesma need to filter out based on it when rendering hits

Therefore when fetching document with specific `_id` we could filter out documents with matching timestamp and then filter for that with matching fields, returning the right entry.

This fixes the issue, where JSON view of single document could return a random object from search hits, not necessarily the one clicked.
image

**However** (and it's a pretty big however), the "surrounding document" view **cannot return the surrounding documents**. While the query doesn't error it also doesn't return any documents. The experiment to have that working [is carried out is a separate PR](https://github.com/QuesmaOrg/quesma/pull/1446), although it's not clear yet if this way of doing things is going to guarantee 100% correctness.


platform/model/typical_queries/hits.go
```diff
@@ -241,7 +244,8 @@ func (query Hits) computeIdForDocument(doc model.SearchHit, defaultID string) st
 			// At database level we only compare timestamps with millisecond precision
 			// However in search results we append `q` plus generated digits (we use q because it's not in hex)
 			// so that kibana can iterate over documents in UI
-			pseudoUniqueId = fmt.Sprintf("%xq%s", vv, defaultID)
+			sourceHash := fmt.Sprintf("%x", ComputeHash(doc.Source))
+			pseudoUniqueId = fmt.Sprintf("%xqqq%sqqq%x", vv, defaultID, sourceHash)
```
jakozaur, line 248: What is this `defaultID`?

platform/model/typical_queries/hits.go
```diff
@@ -250,6 +254,44 @@ func (query Hits) computeIdForDocument(doc model.SearchHit, defaultID string) st
...
+	if err != nil {
+		hash := sha256.Sum256(data)
+		return hex.EncodeToString(hash[:])
+	}
+	hash := sha256.Sum256(normalizedBytes)
+	return hex.EncodeToString(hash[:])
+}
+
+// normalizeJSON recursively normalizes JSON structure to ensure consistent ordering for further hashing.
+func normalizeJSON(v interface{}) interface{} {
```
jakozaur, line 274: I do not understand how this guarantees consistent ordering. map have no guarantee of ordering. In particular entering same objects to map with a different order may result in a different map.

platform/parsers/elastic_query_dsl/query_translator.go
```diff
@@ -24,7 +26,8 @@ type ClickhouseQueryTranslator struct {
 	Indexes []string
 
 	// TODO this will be removed
-	Table *clickhouse.Table
+	Table     *clickhouse.Table
+	UniqueIDs []string // used for hits queries, to filter out hits that are not in the list of IDs
```
jakozaur, line 30: Maybe a better name? Filter unique IDs at the end?

platform/model/typical_queries/hits.go
jakozaur: Shouldn't we delete it.

jakozaur (approved):
Some comments.

platform/parsers/elastic_query_dsl/query_translator.go
mieciu: how about https://github.com/QuesmaOrg/quesma/pull/1435/commits/80dd2667e4ee0de59d99d403942edf471646d836 ?

platform/model/typical_queries/hits.go
mieciu: https://github.com/QuesmaOrg/quesma/pull/1435/commits/e8cc635cc7645e090e522f3393f760b8f21753dd

jakozaur: Ok, will still moving to next doc work after your change?

platform/parsers/elastic_query_dsl/query_translator.go
jakozaur: Great!

jakozaur (approved):

PR #1439, opened 2025-06-04
Refactor towards better table creation

jakozaur:
Our logic of parsing `CREATE TABLE` needs to be replaced. This is code duplication and a potential issue.

First steps:
1. Add logic to log proper bugs instead of eating them silently.
2. Add some tests that reproduce the client issue. Though we decided against fixing our parser.
3. Start by removing the confusing field `Created`. It is not used now, it is artifact from legacy hardcoding table.


nablaone (approved):
LGTM

pdelewski:
@jakozaur That's the first step?, if I understand correctly

jakozaur:
@pdelewski Yes, there is more refactoring. Though decided step by step approach is better.

PR #1441, opened 2025-06-04
Limit logging errors on ingest

nablaone:
This PR fixes logging errors of failed ingest statements:
1. We should log the root cause.
2. We should limit the amount of logging.


jakozaur (approved):

pdelewski:
I would suggest to extract these hardcoded values as copilot suggest

PR #1442, opened 2025-06-05
Dont expose internal table/index

nablaone:
Users should not be able to create a data view that matches the common table.  Querying the common table causes an error. 

This PR filters out our internal common table and index that contains a list of virtual tables.

### DIFF
--- a/platform/functionality/resolve/resolve.go
+++ b/platform/functionality/resolve/resolve.go
@@ -24,7 +24,7 @@ func HandleResolve(pattern string, sr schema.Registry, ir elasticsearch.IndexRes
 		sourcesToShow = &sourcesFromElasticsearch
 	}
 
-	var filteredOut []elasticsearch.Index
+	var filtered []elasticsearch.Index
 	for _, index := range sourcesToShow.Indices {
 		if index.Name == common_table.VirtualTableElasticIndexName {
 			// don't include the common table in the results
--- a/platform/functionality/resolve/resolve.go
+++ b/platform/functionality/resolve/resolve.go
@@ -3,6 +3,7 @@
 package resolve
 
 import (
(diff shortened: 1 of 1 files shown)


platform/functionality/resolve/resolve.go
```diff
@@ -23,16 +24,28 @@ func HandleResolve(pattern string, sr schema.Registry, ir elasticsearch.IndexRes
 		sourcesToShow = &sourcesFromElasticsearch
 	}
 
+	var filteredOut []elasticsearch.Index
```
mieciu, line 27:
```suggestion
	var filtered []elasticsearch.Index
```
`filteredOut` is a little confusing, because this array contains items that are actually _kept_

platform/functionality/resolve/resolve.go
```diff
@@ -23,16 +24,28 @@ func HandleResolve(pattern string, sr schema.Registry, ir elasticsearch.IndexRes
...
+	for _, index := range sourcesToShow.Indices {
+		if index.Name == common_table.VirtualTableElasticIndexName {
```
mieciu, line 29:
```suggestion
 		// don't include the common table in the results - it's internal table used by Quesma, and should be exposed as an index / data stream
		if index.Name != common_table.VirtualTableElasticIndexName {
		        filtered = append(filtered, index)
 		}
 		
```

mieciu: I think this is simpler to read rather than `continue`, but may be just my preference :)

mieciu (commented):
Just random readability notes, other than that -> 💪  

LMK if you disagree, in case of which I'm good to merge as is

mieciu (approved):
:shipit:

PR #1443, opened 2025-06-05
Remove most of parse create table

jakozaur:
This is stage 2 of removing the parse create table. Not completely, and still no logic to just rely on schema, but it updates all tests as well as vastly reduces the scope of the create table parser.

As a side-effect, it bugfixes tables with dots in the names.


cmd/v2_test_objects.go
```diff
@@ -388,17 +388,19 @@ func (p *QueryTransformationPipeline) ParseQuery(message any) (*model.ExecutionP
...
 	// Why parser needs a table?
 	tableName := "test_table"
-	table, err := clickhouse.NewTable(`CREATE TABLE `+tableName+`
-		( "message" String, "@timestamp" DateTime64(3, 'UTC'), "attributes_values" Map(String,String))
-		ENGINE = Memory`,
-		clickhouse.NewNoTimestampOnlyStringAttrCHConfig(),
-	)
-	if err != nil {
-		return nil, err
+	table := clickhouse.Table{
```
mieciu, line 391: I still like the idea of having this behind constructor, esp. given that this repeats in few places.

Of course I'd make `clickhouse.NewTable()` take table name and list of columns not the `CREATE TABLE ...` string 😉

jakozaur:
> A very much needed refactor, great job!
> 
> I'd still prefer to have `clickhouse.Table` object not being initialized directly, but not going to push back because of that.

I will, though I’ll probably merge it as-is. I’ve already significantly expanded the scope of my work just to fix a tiny bug. But instead of adding more duct tape (fix table names with dots), I chose a major simplification approach (don't rely on parsing of create tables).

PR #1455, opened 2025-06-10
Add `split_time_range` integration test

avelanarius:
Recently, there was a regression caused by a bug in `splitTimeRangeExt` optimization (see #1454), which splits queries with long time ranges. Our unit/integration tests failed to catch the issue. The reproducer for the issue is to run a `_search` query: sorted by time, over some time range and with a filter.

This PR adds a new integration test. This test runs the reproducing query (`_search` query: sorted by time, over some time range and with a filter), as well as different variations of the query: short/long time range, with/without filter, inclusive/exclusive ends of time range. This should cover the `splitTimeRangeExt` optimization more thoroughly.

There are two types of tests:
- `testTimerangesHandwritten`: handwritten different queries
- `testTimerangesRandom`: randomly generated ranges

### DIFF
--- a/ci/it/configs/quesma-split-time-range.yml.template
+++ b/ci/it/configs/quesma-split-time-range.yml.template
@@ -1,0 +1,57 @@
+frontendConnectors:
+  - name: elastic-ingest
+    type: elasticsearch-fe-ingest
+    config:
+      listenPort: 8080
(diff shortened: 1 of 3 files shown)


nablaone (approved):
LGTM

PR #1460, opened 2025-06-11
[draft]Use Apache Doris as the data connection source

DongLiang-0:
A draft PR using quesma to use doris as the data connection source.
This connection source is modified based on clickhouse


examples/kibana-sample-data/docker-compose.yml
```diff
@@ -1,33 +1,33 @@
 services:
```
jakozaur, line 1: Ideally, you can copy this file and create another like `examples/doris-kibana-sample-data` for your use case.

ci/ci.yml
```diff
@@ -18,7 +18,7 @@ services:
       - "9200:8080"
     volumes:
       - ./quesma/logs/:/var/quesma/logs
-      - ../examples/kibana-sample-data/quesma/config/local-dev.yaml:/mnt/ci-config.yaml
+      - ../examples/kibana-sample-data/quesma/config/config.yaml:/mnt/ci-config.yaml
```
jakozaur, line 21: Will see, but probably adding another file would be easier than modifying it.

examples/kibana-sample-data/docker-compose.yml
```diff
@@ -50,11 +50,11 @@ services:
   kibana:
     image: docker.elastic.co/kibana/kibana:8.11.1
     environment:
-      ELASTICSEARCH_HOSTS: '["http://quesma:8080"]'
+      ELASTICSEARCH_HOSTS: '["http://10.16.10.6:8080"]'
```
jakozaur, line 53: If you like custom ip addresses, you may add another folder in `examples/` such as `examples/dev-...` and then add it to `.gitignore`.

platform/backend_connectors/basic_sql_backend_connector.go
```diff
@@ -72,10 +74,12 @@ func (p *BasicSqlBackendConnector) Query(ctx context.Context, query string, args
 }
 
 func (p *BasicSqlBackendConnector) QueryRow(ctx context.Context, query string, args ...interface{}) quesma_api.Row {
+	logger.Info().Msgf("queryRow sql: %s", query)
 	return p.connection.QueryRowContext(ctx, query, args...)
 }
 
 func (p *BasicSqlBackendConnector) Exec(ctx context.Context, query string, args ...interface{}) error {
+	logger.Info().Msgf("Exec sql: %s", query)
```
jakozaur, line 82: All three above make a huge sense for debugging. You may also leverage our internal debug console `http://localhost:9999`.

jakozaur (commented):
Amazing proof of concept. Thank you for starting it.

Though to merge it to `main` we would probably need:
1. Work on new backend connector. Start by copying `platform/clickhouse` to `platform/doris` and make changes there, while avoid changes that would break ClickHouse.
2. That would probably need also some changes to design.
3. Similarly with SQL function changes. Ideally we should build abstract syntax tree and render the SQL dialects in `expr_string_renderer.go`
4. Make sure you will be able to sign CLA.

platform/backend_connectors/basic_sql_backend_connector.go
DongLiang-0: thx

platform/clickhouse/connection.go
```diff
@@ -6,9 +6,7 @@ import (
 	"crypto/tls"
```
pdelewski, line 6: I would suggest to create a new directory `apache_doris` and implement below functionality there. Then proper `initDBConnection` can be invoked here https://github.com/QuesmaOrg/quesma/blob/main/cmd/main.go#L87

pdelewski:
> Hey @pdelewski , the two problems you mentioned above are not difficult. Now the main problem is another one: all the parsing and transform statements in the framework are for ClickHouse(there are some syntax differences between doris and clickhouse). If we forcefully add Doris-related codes, the code may be confusing.

@DongLiang-0 

> Hey @pdelewski , the two problems you mentioned above are not difficult. Now the main problem is another one: all the parsing and transform statements in the framework are for ClickHouse(there are some syntax differences between doris and clickhouse). If we forcefully add Doris-related codes, the code may be confusing.

@DongLiang-0 True, but there are at least few options to solve this problem. I will add some comments in different places.

platform/parsers/elastic_query_dsl/pancake_sql_query_generation.go
```diff
@@ -51,19 +65,41 @@ func (p *pancakeSqlQueryGenerator) generateAccumAggrFunctions(origExpr model.Exp
...
+		case "NDV":
+			firstArg := origFunc.Args[0]
+			var columnName string
+			if colRef, ok := firstArg.(model.ColumnRef); ok {
+				columnName = colRef.ColumnName
+			}
+			return model.NewFunction("APPROX_COUNT_DISTINCT", model.NewColumnRef(columnName)), "", nil
 		}
 
-		if strings.HasPrefix(origFunc.Name, "quantiles") {
```
pdelewski, line 64: Same as above, Do we need to remove that?

platform/parsers/elastic_query_dsl/pancake_sql_query_generation.go
```diff
@@ -37,10 +45,16 @@ func (p *pancakeSqlQueryGenerator) aliasedExprArrayToLiteralExpr(aliasedExprs []
 	return exprs
 }
 
-func (p *pancakeSqlQueryGenerator) generatePartitionBy(groupByColumns []model.AliasedExpr) []model.Expr {
+func (p *pancakeSqlQueryGenerator) generatePartitionBy(groupByColumns []model.GroupByExpr, useGroupByColumn bool) []model.Expr {
```
pdelewski, line 48: I would suggest to rename old function to something like `generatePartitionByViaAliasExpr`  and add a new function `generatePartitionByViaGroupByExpr`

platform/parsers/elastic_query_dsl/pancake_sql_query_generation.go
```diff
@@ -89,55 +125,59 @@ func (p *pancakeSqlQueryGenerator) generateMetricSelects(metric *pancakeModelMet
...
+			if alias.GroupAlias == orderByLiteral.Value {
 				return true
 			}
 		}
 	}
 	return false
 }
 
-func (p *pancakeSqlQueryGenerator) addPotentialParentCount(bucketAggregation *pancakeModelBucketAggregation, groupByColumns []model.AliasedExpr) []model.AliasedExpr {
+func (p *pancakeSqlQueryGenerator) addPotentialParentCount(bucketAggregation *pancakeModelBucketAggregation, groupByColumns []model.GroupByExpr) []model.AliasedExpr {
```
pdelewski, line 160: Same as above

platform/model/expr.go
```diff
@@ -202,6 +202,19 @@ func (t TableRef) Accept(v ExprVisitor) interface{} {
 	return v.VisitTableRef(t)
 }
 
+type GroupByExpr struct {
```
pdelewski, line 205: I don't understand the difference between the new GroupByExpr and the existing AliasExpr types — their contents appear to be the same. Introducing this new type triggers many changes in function signatures, and I'm not sure why it's needed.

pdelewski: I understand that, however now `GroupByExpr` and `AliasExpr ` are the same. Do you plan to change that?

PR #1461, opened 2025-06-12
Lack of schema refresh after new field in common table

nablaone:
Quesma relies on `attributes` columns when it calculates the schema change on ingest. Table discovery didn't add these fields to virtual tables.  Therefore, the schema of the index stored in the `common table` becomes immutable within 60 seconds.

### DIFF
--- a/ci/it/testcases/test_only_common_table.go
+++ b/ci/it/testcases/test_only_common_table.go
@@ -9,7 +9,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"net/http"
 	"testing"
-	"time"
 )
 
 type OnlyCommonTableTestcase struct {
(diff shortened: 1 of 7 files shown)


ci/it/testcases/test_only_common_table.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (a *OnlyCommonTableTestcase) testAlterVirtualTable(ctx context.Context, t *testing.T) {
+
+	resp, body := a.RequestToQuesma(ctx, t, "POST", "/logs-6/_doc", []byte(`{"name": "Przemyslaw", "age": 31337}`))
+	fmt.Println(string(body))
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// wait for internal processing, especially for the periodic task that updates the schema
+	time.Sleep(60 * time.Second)
```
jakozaur, line 46: Ok, this is very tricky to add. Slow IT are hard to debug and takes a lot to some engineering.

jakozaur (commented):
Test should be faster.

ci/it/testcases/test_only_common_table.go
```diff
@@ -0,0 +1,67 @@
...
+func (a *OnlyCommonTableTestcase) RunTests(ctx context.Context, t *testing.T) error {
+
+	t.Run("test alter virtual table", func(t *testing.T) { a.testAlterVirtualTable(ctx, t) })
+	return nil
+}
+
+func (a *OnlyCommonTableTestcase) testAlterVirtualTable(ctx context.Context, t *testing.T) {
+
+	reloadTables := func() {
+		resp, body := a.RequestToQuesma(ctx, t, "POST", "/_quesma/reload-tables", nil)
```
jakozaur, line 41: We should put that method under our admin port, not on ElasticSearch frontend.

As we are rushing with release fine to do in next PR.

platform/frontend_connectors/router_v2.go
```diff
@@ -310,5 +310,16 @@ func ConfigureSearchRouterV2(cfg *config.QuesmaConfiguration, dependencies quesm
 		return &quesma_api.Result{Body: string(body), StatusCode: http.StatusOK, GenericResult: body}, nil
 	})
 
+	router.Register(routes.QuesmaReloadTablsPath, method("POST"), func(ctx context.Context, req *quesma_api.Request, _ http.ResponseWriter) (*quesma_api.Result, error) {
```
jakozaur, line 313: As above, please move this to admin interface.

jakozaur (approved):
Can be merged, but move it to admin interface.

PR #1462, opened 2025-06-12
This fixes Kibana flights sample dashboards, it also makes logs less noisy

jakozaur:
Kibana flights dashboards was broken as Carrier should be `keyword` not `text`. More systematical change is harder.

Also logs end up being very noisy and confusing. End up cleaning them.


nablaone (commented):
LGTM

nablaone (approved):

PR #1468, opened 2025-06-17
add apache doris docker environment

DongLiang-0:
Add a docker environment of apache doris


pdelewski:
@DongLiang-0 After yesterday discussion, we got to conclusion that it would be better to have separate docker file per each backend environment, e.g not to have `clickhouse` and `doris` in one environment.

pdelewski:
@DongLiang-0 One more thing — as mentioned earlier, you need to accept the CLA before we can merge your contribution.

PR #1472, opened 2025-06-18
Exctracting ingest ddl transformation

pdelewski:
This PR extracts and introduces a new transformation interface:

```
type Lowerer interface {
    LowerToDDL(validatedJsons []types.JSON,
        table *chLib.Table,
        invalidJsons []types.JSON,
        encodings map[schema.FieldEncodingKey]schema.EncodedFieldName,
        createTableCmd CreateTableStatement) ([]string, error)
}
```
The intention is to transpile input into other DDL representations, with the primary goal of supporting Hydrolix ingestion.

It also introduces basic abstractions for building DDL using an object model instead of raw strings.

This interface may evolve over time to adapt to our needs.

### DIFF
--- a/platform/ingest/parser.go
+++ b/platform/ingest/parser.go
@@ -67,6 +67,10 @@ func columnsToProperties(columnsFromJson []CreateTableEntry,
 		if columnFromSchema, found := columnsFromSchema[schema.FieldName(columnFromJson.ClickHouseColumnName)]; found {
 			// Check if the type is an Array – if so, fallback to JSON type
 			if strings.Contains(columnFromJson.ClickHouseType, "Array") {
+				// The schema (e.g. PUT /:index/_mapping) doesn't contain information about whether a field is an array or not.
+				// Therefore, we have to combine the information from the schema and the JSON in such case.
+				// For example: in the mapping we have a field "products.name" with type "keyword" (String)
+				// and in the JSON "products.name" is an array of strings (Array(String)).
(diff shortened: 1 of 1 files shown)


platform/ingest/parser.go
```diff
@@ -31,72 +31,105 @@ func reverseFieldEncoding(fieldEncodings map[schema.FieldEncodingKey]schema.Enco
...
+	}
+}
+
+// Returns a slice of ColumnProperties containing all needed column properties
+func columnsToProperties(columnsFromJson []CreateTableEntry,
 	columnsFromSchema map[schema.FieldName]CreateTableEntry,
 	fieldEncodings map[schema.FieldEncodingKey]schema.EncodedFieldName,
 	tableName string,
-) string {
+) []ColumnStatement {
```
mieciu, line 55: perfect!

platform/ingest/ddllowerer.go
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Elastic-2.0
+package ingest
+
+import (
+	chLib "github.com/QuesmaOrg/quesma/platform/clickhouse"
+	"github.com/QuesmaOrg/quesma/platform/schema"
+	"github.com/QuesmaOrg/quesma/platform/types"
+)
+
+type Lowerer interface {
```
mieciu, line 11: I would prefer to see the PR description here over the interface declaration :)

PR #1477, opened 2025-06-24
Log less when elastic is down

nablaone:
This PR limits error logs when elastic is not available.


platform/frontend_connectors/dispatcher.go
```diff
@@ -495,8 +499,23 @@ func (r *Dispatcher) sendHttpRequest(ctx context.Context, address string, origin
 
 	resp, err := r.HttpClient.Do(req)
 	if err != nil {
-		logger.ErrorWithCtxAndReason(ctx, "No network connection").
-			Msgf("Error sending request: %v", err)
+
+		var shouldLogError bool
```
pdelewski, line 503: Maybe
```suggestion
shouldLogError := true

// Rate-limit only network-related errors (e.g., dial issues)
if strings.Contains(err.Error(), "dial") {
	if r.elasticSearchErrorRateLimiter != nil && !r.elasticSearchErrorRateLimiter.Allow() {
		shouldLogError = false
	}
}
```

pdelewski (approved):

PR #1482, opened 2025-07-01
[draft]support doris connector

DongLiang-0:


platform/clickhouse/table.go
```diff
@@ -101,10 +101,10 @@ func (t *Table) GetDateTimeType(ctx context.Context, fieldName string, dateInSch
 	if col, ok := t.Cols[fieldName]; ok {
 		typeName := col.Type.String()
 		// hasPrefix, not equal, because we can have DateTime64(3) and we want to catch it
-		if strings.HasPrefix(typeName, "DateTime64") {
+		if strings.HasPrefix(typeName, "datetime") {
```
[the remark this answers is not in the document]
pdelewski: Let's try this way

platform/clickhouse/table.go
```diff
@@ -101,10 +101,10 @@ func (t *Table) GetDateTimeType(ctx context.Context, fieldName string, dateInSch
...
 			return DateTime64
 		}
-		if strings.HasPrefix(typeName, "DateTime") {
+		if strings.HasPrefix(typeName, "date") {
```
pdelewski, line 107:
```suggestion
		if strings.HasPrefix(typeName, "DateTime") || strings.HasPrefix(typeName, "date") {
```

platform/clickhouse/schema.go
```diff
@@ -220,35 +221,75 @@ func NewBaseType(clickHouseTypeName string) BaseType {
 // this is catch all type for all types we do not exlicitly support
 type UnknownType struct{}
 
-func ResolveType(clickHouseTypeName string) reflect.Type {
-	switch clickHouseTypeName {
-	case "String", "LowCardinality(String)", "UUID", "FixedString":
+func ResolveType(dorisTypeName string) reflect.Type {
```
pdelewski, line 224: I think we have to have type resolver per each backend. It would make sense to introduce 
```
type TypeResolver interface {
  ResolveType(typeName string) reflect.Type
}
```

PR #1483, opened 2025-07-02
Backend support refactoring

pdelewski:


pdelewski:
Closing for now as we tool a bit simpler approach in https://github.com/QuesmaOrg/quesma/pull/1485

PR #1491, opened 2025-07-07
Regression fix for `CREATE TABLE ... ON CLUSTER`

mieciu:
We have recently introduced a regression which could result in table creation failures whenever `ON CLUSTER` clause had to be applied.


platform/ingest/insert_test.go
```diff
@@ -33,64 +31,41 @@ import (
...
-			`	"message" String`,
-			`	"random1" Array(String)`,
-			`	"random2" string`,
-			`	"severity" String`,
-			`	INDEX severity_idx severity TYPE set(25) GRANULARITY 4`,
-			`	`,
-			``,
-			`)`,
+			`CREATE TABLE IF NOT EXISTS "test_table" `,
+			//`	"@timestamp" DateTime64`,
```
pdelewski, line 59: Why is this commented out?

PR #1495, opened 2025-07-16
Adapt doris schema type

DongLiang-0:

### DIFF
--- a/platform/database_common/schema.go
+++ b/platform/database_common/schema.go
@@ -259,7 +259,7 @@ func GetTypeResolver(instanceType InstanceType) TypeResolver {
 	case ClickHouseInstance:
 		r = &clickhouse.ClickhouseTypeResolver{}
 	default:
-		logger.Fatal().Msgf("unknown instance type: %s", instanceType)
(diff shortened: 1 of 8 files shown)


platform/database_common/schema.go
```diff
@@ -211,44 +212,38 @@ func (t MultiValueType) GetColumn(name string) *Column {
...
-		return reflect.TypeOf(clickhouse.Point{})
-	case "Bool":
-		return reflect.TypeOf(true)
-	case "JSON":
-		return reflect.TypeOf(map[string]interface{}{})
-	case "Map(String, Nullable(String))", "Map(String, String)", "Map(LowCardinality(String), String)", "Map(LowCardinality(String), Nullable(String))":
-		return reflect.TypeOf(map[string]string{})
-	case "Unknown":
-		return reflect.TypeOf(UnknownType{})
+func NewBaseTypeWithInstanceName(typeName string, instanceName string) BaseType {
```
pdelewski, line 225: I think it would be better to use some const instead of string as `instanceName`

pdelewski (approved):

PR #1500, opened 2025-07-22
Use backquote for table and column names

DongLiang-0:
Previously, columns and tables in SQL used double quotes, but in doris syntax, columns and tables should use back quotes, otherwise they will be recognized as a string. This change is applicable to both clickhouse and doris
```
select "col1" from tab1 -> select `col1` from `tab1`
```


mieciu (changes requested):
Thanks a lot for your contribution!

I think that instead of swapping " with `, this should be **configurable, depending on the backend connector.** While this syntax is fine with ClickHouse/Hydrolix, it's non supported in Postgres. 

Therefore, there should be no need to change any of the existing test cases.

### How I would see this change moving forward

We have this oversimplification of `model.LiteralExpr` being used for column/table names and aliases. This is the moment where we need to refactor that.

We need to create `model.IdentifierExpr` for our query AST which should replace `model.LiteralExpr` whenever referring to column/table name. Then, **when rendering to SQL** we should transform it accordingly to the database backend we're connected to. ClickHouse/Hydrolix -> let's stick to `"`. For Doris, we can use "`". 

In that way, we won't have to change any of the existing tests, make adding different data sources easier and there's way less risk of introducing regression to the codebase.

DongLiang-0:
I'll close this PR.

PR #1501, opened 2025-07-23
Adapt to string type time data

DongLiang-0:
Since the data is converted into string type in the data reading of doris, we need to process the string type specifically here.

Related PR: https://github.com/QuesmaOrg/quesma/blob/main/platform/database_common/quesma_communicator.go#L228

### DIFF
--- a/platform/model/bucket_aggregations/date_histogram.go
+++ b/platform/model/bucket_aggregations/date_histogram.go
@@ -222,21 +222,7 @@ func (query *DateHistogram) generateSQLForCalendarInterval() model.Expr {
 }
 
 func (query *DateHistogram) getKey(row model.QueryResultRow) int64 {
-	value := row.Cols[len(row.Cols)-2].Value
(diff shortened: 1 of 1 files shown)


platform/model/bucket_aggregations/date_histogram.go
```diff
@@ -222,7 +222,21 @@ func (query *DateHistogram) generateSQLForCalendarInterval() model.Expr {
 }
 
 func (query *DateHistogram) getKey(row model.QueryResultRow) int64 {
-	return row.Cols[len(row.Cols)-2].Value.(int64)
+	value := row.Cols[len(row.Cols)-2].Value
```
pdelewski, line 225: Seems that these 2 functions are the same. Please extract one helper function and use it in both places

pdelewski (changes requested):
Seems that these 2 functions are the same. Please extract one helper function and use it in both places

PR #1507, opened 2025-08-06
Log column name when validation failed

mieciu:
It is hard to determine which part of the payload (or even ingesting service) is ingesting incorrect data based on just:
```
Invalid value type for column of type Int64: string, value: <SOME_RANDOM_STRING>
```
Putting column name would make tracing the source of the problem so much easier.


nablaone (approved):
LGTM

